PR #76, opened 2019-05-10
fixed black checkstyle

chriskfwoo:
Currently, the master branch does not pass the black checkstyle.


apps/dash-pk-calc/app.py
```diff
@@ -20,163 +20,273 @@
...
-                       'conc': [1.1, 3.04, 4.85, 3.93, 2.01, 1.02, .51, 0.25,
-                                0.92, 2.8, 4.6, 4.1, 1.99, 1.05, .55, 0.3,
-                                1.04, 3.23, 5.35, 4.1, 2.4, 1.12, .52, 0.27]
-                       })
+pkdata = pd.DataFrame(
+    {
+        "subject_index": [
+            0,
+            0,
+            0,
```
chriskfwoo: I applied the black linter. I'm not sure if this makes it easier to read.
Is there a way to ignore long list of arrays?

bpostlethwaite: Put them in a CSV/JSON/Whatever file and read them into Python when the app starts :) 
Otherwise nope. Black is very opinionated.

shammamah: @chriskfwoo I'll take over here -- I put this app in without checking to see if the CI tests passed.

PR #82, opened 2019-05-15
Fix gallery upload script to only look in apps/ directory.

shammamah:
Previously, in the script, `.github/ISSUE_TEMPLATE` would count as part of the diff during the gallery deployment, which caused it to fail.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
       - run:
           name: Deploy
           command: |
-            APPS_MODIFIED=$(git diff origin/production origin/master --dirstat=files,1 | awk '{ $2 ~ /apps/; split($2,a,"/"); if (length(a[2]) != 0) { print a[2]} }')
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -52,7 +52,7 @@ jobs:
       - run:
           name: Deploy
           command: |
-            APPS_MODIFIED=$(git diff origin/production origin/master --dirstat=files,1 | awk '{ $2 ~ /apps/; split($2,a,"/"); if (length(a[2]) != 0) { print a[2]} }')
+            APPS_MODIFIED=$(git diff origin/production origin/master --dirstat=files,1 apps/ | awk '{ $2 ~ /apps/; split($2,a,"/"); if (length(a[2]) != 0) { print a[2]} }' | sort -u)
             for APP in $APPS_MODIFIED
```
tarzzz: You can actually simplify this even more:
```
git diff origin/production master --dirstat=files,1 apps/ | awk '{ split($2,a,"/"); if(length(a[2]) != 0) { print a[2]} }' | sort -u
```

PR #83, opened 2019-05-15
Update predeploy

rpkyle:
This PR proposes changes to enable R app deployments, including addition of `Aptfile` and `init.R`, as well as minor edits to `predeploy.py`.

### DIFF
--- a/predeploy.py
+++ b/predeploy.py
@@ -6,14 +6,11 @@ from apps_directory_mapping import APPNAME_TO_DIRECTORY
 
 
 files = ["requirements.txt", "Procfile"]
-rfiles = ["Aptfile", "init.R"] 
+rfiles = ["Aptfile", "init.R"]
 
 app_name = os.environ["DASH_APP_NAME"]
 
-app_path = os.path.join(
-    "apps",
-    APPNAME_TO_DIRECTORY.get(app_name, app_name),
-)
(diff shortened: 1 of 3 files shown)


predeploy.py
```diff
@@ -6,16 +6,20 @@
...
     shutil.copyfile(os.path.join(app_path, f), f)
 
 if "DOKKU_SCALE" in os.listdir(app_path):
     shutil.copyfile(os.path.join(app_path, "DOKKU_SCALE"), "DOKKU_SCALE")
 
+if "-" in app_name and app_name.split("-")[0] == "dashr":
+    for rfile in rfiles:
+        shutil.copyfile(os.path.join(app_path, rfile), rfile)
+
 subprocess.run("python -m pip install -r requirements.txt".split(" "))
```
tarzzz: @rpkyle for `R`, this last line will fail, so might wanna put it inside a condition so that it is only executed for python apps:

```
if "-" in app_name and app_name.split("-")[0] == "dashr":
    for rfile in rfiles:
        shutil.copyfile(os.path.join(app_path, rfile), rfile)
else:
    subprocess.run("python -m pip install -r requirements.txt".split(" "))
```

This is not really a very clean way IMHO, but it should do until we refactor `predeploy.py` to have different code-paths for `R` and `python` apps.

PR #86, opened 2019-05-15
Circle CI auto-deploy improvements

tarzzz:


tarzzz:
Some updates:
  - If you don't change an app, we do not attempt to deploy to playground and the tests pass.
  - If you change an app, but do not use the same branch-name.. the tests fail and the app is _not_ deployed to playground.

tarzzz:
@shammamah Please review..

Issue #37, opened 2019-05-01
Interactive t-SNE [PYTHON]


shammamah:
# App name
dash-tsne

# App
- [x] This app is currently deployed on Dash(R) Gallery: https://dash-gallery.plotly.host/dash-tsne/
- [ ] This app is new

# Source code repository
https://github.com/plotly/dash-tsne

# Current appearance
![2019-05-01 13 36 48 dash-gallery plotly host 148790289973](https://user-images.githubusercontent.com/15622051/57031624-2eb1df00-6c16-11e9-993b-fd38a94b444e.png)

# Redesign
Link: https://github.com/plotly/streambed/issues/12068#issuecomment-461590032

![](https://user-images.githubusercontent.com/37411533/52445378-f019f600-2af8-11e9-8721-e9105bcb3e12.png)

Link: https://github.com/plotly/streambed/issues/12879#issuecomment-488090333

![](https://user-images.githubusercontent.com/37411533/56989020-6dd12900-6b5f-11e9-9c38-36855b87f85c.jpg)

# Playground deployment 
[url of playground deployment]

PR #94, opened 2019-05-17
dash-tsne

celinehuang:
# App pull request
- [ ] This is a new app

## About

- Playground deployment URL (new version): https://dash-playground.plotly.host/dash-tsne
- Current gallery app URL: https://dash-gallery.plotly.host/dash-tsne/
- Streambed issue card with link to redesign mockup: https://github.com/plotly/dash-sample-apps/issues/37

## Workflow

image

## The pre-review review


shammamah:
@celinehuang looking at the deployed version on playground now, and it doesn't really look like the mock-up. (Assets path might also be messed up, since the Dash logo isn't loading properly). Please fix these issues and I can review!

celinehuang:
@shammamah Fixed the assets path; I switched the logo and title positions in the overall layout in order for the mobile layout to have a centered title, hope that's ok!

shammamah:
@celinehuang Let's try to get it as close as possible to the mobile layout. In the web layout, you can probably get away with absolute positioning for the header and the logo, or change the widths of the logo and title to 100% once you get into mobile layout (and then they will stack).

The colour of the header also looks a little off to me in the web layout. I know these are seemingly minor details, but it's good to get it as close as possible to the mockup. If you find that this is a blocking issue I'm happy to help!

shammamah:
(there's also a reference to "the repository" (for people who want to upload their own datasets). Not sure if we should keep this, since those repos will be archived.

apps/dash-tsne/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
shammamah: We don't need this LICENSE file.

apps/dash-tsne/README.md
```diff
@@ -0,0 +1,97 @@
...
+- **Celine Huang** [@celinehuang](https://github.com/celinehuang)
+- **Xing Han Lu** - _Initial Work_ - [@xhlulu](https://github.com/xhlulu)
+
+See also the list of [contributors](https://github.com/plotly/dash-svm/contributors) who participated in this project.
+
+## License
+
+This project is licensed under the MIT License - see the [LICENSE.md](LICENSE.md) file for details
+
+## Acknowledgments
```
ronniebugia: you can just get rid of this if there are no Acknowledgements.

apps/dash-tsne/assets/styles.css
```diff
@@ -0,0 +1,200 @@
...
+    padding-right: 7%;
+  }
+
+  #learn-more-button > div > div:nth-child(3) {
+    padding-right: 15%;
+  }
+}
+@media only screen and (max-width: 550px) {
+  #plotly-image {
+    height: 35px !important;
```
ronniebugia: It's generally not advisable to put `!important` in css unless necessary. I think for some examples, the css call is so specific that `!important` isn't needed.

apps/dash-tsne/assets/styles.css
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  #app-title {
+    text-align: center !important;
+  }
+
+  .header_img {
+    height: 40px;
+    float: left;
+    margin: 0 !important;
```
ronniebugia: Setting `margin: 0 !important` isn't necessary here. I would try getting rid of as many `!important` calls as possible.

apps/dash-tsne/demo.py
```diff
@@ -0,0 +1,676 @@
...
+        style=merge(
+            {
+                "padding": 20,
+                "margin": 5,
+                "borderRadius": 5,
+                "user-select": "none",
+                "-moz-user-select": "none",
+                "-webkit-user-select": "none",
+                "-ms-user-select": "none",
+            },
```
ronniebugia: I would put all this as a class name in your `css` instead of declaring inline.

apps/dash-tsne/demo.py
```diff
@@ -0,0 +1,624 @@
...
+import io
+import pathlib
+
+import numpy as np
+import dash
+import dash_core_components as dcc
+import dash_html_components as html
+from PIL import Image
+from io import BytesIO
+from dash.dependencies import Input, Output, State
```
ronniebugia: You never use `State` at all so you can get rid of it.

apps/dash-tsne/demo.py
```diff
@@ -0,0 +1,624 @@
...
+                if not selected_word:
+                    return go.Figure()
+
+                plot_mode = "text"
+
+                # Get the nearest neighbors indices using Euclidean distance
+                vector = data_dict[dataset].set_index("0")
+                selected_vec = vector.loc[selected_word]
+
+                def compare_pd(vector):
```
ronniebugia: Why is `compare_pd` not declared outside of this function. Also, `compare_pd` is also used in line 576 and line 579. So I recommend just declaring `compare_pd` outside of this function's scope.

apps/dash-tsne/demo.py
```diff
@@ -0,0 +1,624 @@
...
+            figure = go.Figure(data=[scatter], layout=layout)
+
+            return figure
+        except KeyError as error:
+            print(error)
+            raise PreventUpdate
+
+    @app.server.before_first_request
+    def load_image_data():
+        global data_dict
```
ronniebugia: Is a `global` variable necessary in this case?

apps/dash-tsne/demo.py
ronniebugia: You never use `State` in your app, so you can get rid of this.

apps/dash-tsne/demo.py
ronniebugia: If it is, feel free to ignore this comment. However, I think for Dash apps we are supposed to avoid using `global` variables and if we need them we should use `dcc.Store`. I think you can just declare `data_dict` without using `global`.

apps/dash-tsne/demo.py
ronniebugia: Ignore this comment. `vector.apply()` takes in a function as a parameter, so the declaration within the function.

Issue #43, opened 2019-05-01
Wind Speed Measurement [PYTHON]


shammamah:
# App name
dash-wind-streaming

# App
- [x] This app is currently deployed on Dash(R) Gallery: https://dash-gallery.plotly.host/dash-wind-stream
- [ ] This app is new

# Source code repository
https://github.com/plotly/dash-wind-streaming

# Current appearance
![2019-05-01 13 46 04 dash-gallery plotly host bfd34b55bb77](https://user-images.githubusercontent.com/15622051/57032090-7edd7100-6c17-11e9-820c-8b4e5b0a2193.png)

# Redesign
Link: https://github.com/plotly/streambed/issues/12068#issuecomment-463430412

![](https://user-images.githubusercontent.com/37411533/52759850-688e2480-2fdb-11e9-993c-d09e64d4c90c.png)

Link: https://github.com/plotly/streambed/issues/12879#issuecomment-488393768

![image](https://user-images.githubusercontent.com/15622051/57783835-d6daa400-76fc-11e9-85a4-f1efd124982c.png)

# Playground deployment 
https://dash-playground.plotly.host/dash-wind-streaming/

chriskfwoo:
Redesigned: 

![image](https://user-images.githubusercontent.com/15094206/57553481-a54c8c00-733c-11e9-8854-ce0314e82a13.png)

PR #122, opened 2019-05-31
mobile view for wind streaming

chriskfwoo:
Playground Link: https://dash-playground.plotly.host/dash-wind-streaming/

Changes
- mobile view for dash-wind-streaming

### DIFF
--- a/apps/dash-wind-streaming/app.py
+++ b/apps/dash-wind-streaming/app.py
@@ -33,7 +33,8 @@ app.layout = html.Div(
                             "This app continually queries a SQL database and displays live charts of wind speed and wind direction.",
                             className="app__header__title--grey",
                         ),
-                    ]
+                    ],
+                    className="app__header__desc",
                 ),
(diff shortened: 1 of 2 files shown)


chriskfwoo:
## Mobile View

wind-mobile

shammamah:
Please ensure that the title is centered like in the mobile mockup here: https://github.com/plotly/dash-sample-apps/issues/43

It might also be a good idea to wrap some of the graphs in `dcc.Loading` components. Moving the slider doesn't seem to result in an immediate change anywhere in the app, which is somewhat confusing.

chriskfwoo:
> Please ensure that the title is centered like in the mobile mockup here: #43
> 
> It might also be a good idea to wrap some of the graphs in `dcc.Loading` components. Moving the slider doesn't seem to result in an immediate change anywhere in the app, which is somewhat confusing.

In my new commit,

image

Added value markers on the slider:
image

PR #137, opened 2019-06-06
Update Graph Interval (dash-wind-streaming) 

### LINKED ISSUE #134: reduce frequency of dash wind streaming requests
# App name
dash-wind-streaming
# Description of bug
At this time this application is making requests very frequently, we could probably improve performance by reducing the frequency of requests while still showing a good example of live streaming 
# Other notes
Suggested by devops team: https://plotly.slack.com/archives/CHLNH6ZTL/p1559755002009900

cc @shammamah to assign out

chriskfwoo:
Closes #134 

Problem
- Sending too many requests

Solution
- Update the interval to 5 seconds by default
- Added an optional environment variable (see [dds env variable](https://dash.plot.ly/dash-deployment-server/environment-variables))


apps/dash-wind-streaming/app.py
```diff
@@ -12,6 +13,8 @@
 from db.api import get_wind_data, get_wind_data_by_id
 
 
+GRAPH_INTERVAL = os.environ.get("GRAPH_INTERVAL", 5000)
```
shammamah: Awesome idea putting this into an environment variable!

PR #144, opened 2019-06-10
Add code to inject Google Analytics script.

### LINKED ISSUE #138: Google Analytics should be part of all apps
It would be best to add the GA code during deployment. 

cc @nicolaskruchten

shammamah:
Closes #138. 

## About 
Added an environment variable in CircleCI with the name `DASH_GA_CODE` that contains the Google Analytics code, and modified the predeploy script to inject that code in an `app.index_string` just above `app.layout` for the app being deployed.  

The script was added into the default `app.index_string` (https://github.com/plotly/dash/blob/7b23e7960f8788fe9050dc2008c99bcc765d0571/dash/dash.py#L40).

### DIFF
--- a/apps/dash-tsne/assets/plotly_ga.css
+++ b/apps/dash-tsne/assets/plotly_ga.css
@@ -1,7 +1,0 @@
-(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
-    
-    ga('create', 'UA-39373211-1', 'auto');
-    ga('send', 'pageview');
--- a/predeploy.py
+++ b/predeploy.py
@@ -21,5 +21,26 @@ if "-" in app_name and app_name.split("-")[0] == "dashr":
 else:
     for f in pyfiles:
         shutil.copyfile(os.path.join(app_path, f), f)
+    lines = []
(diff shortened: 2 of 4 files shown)


shammamah:
@rpkyle I just added this in for the Python deployments -- I think we can do something similar with R, but obviously with some of the variable names (like `app.layout`) changed. Let me know what you think!

predeploy.py
```diff
@@ -21,5 +21,23 @@
...
+    lines = []
+    ga_line = 0
+    with open(os.path.join(app_path, "app.py"), "r") as f:
+        lines = f.readlines()
+
+    try:
+        ga_line = [i for i in range(len(lines)) if "app.layout" in lines[i]]
+        lines.insert(
+            ga_line[0],
+            "app.index_string = '''\n{}\n'''\n\n".format(os.environ["DASH_GA_CODE"]),
```
nicolaskruchten: this is likely not what you want to do, as it will result in no HTML being in `index_string`...

predeploy.py
```diff
@@ -21,5 +21,23 @@
 else:
     for f in pyfiles:
         shutil.copyfile(os.path.join(app_path, f), f)
+    lines = []
+    ga_line = 0
+    with open(os.path.join(app_path, "app.py"), "r") as f:
+        lines = f.readlines()
```
nicolaskruchten: you should be able to move the next block into here and go line by line without slurping it all into memory, by using the iterator.

apps/dash-tsne/assets/plotly_ga.css
```diff
@@ -1,7 +0,0 @@
-(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
```
nicolaskruchten: what is this doing in a CSS file??? :)

shammamah: No idea! 😕

PR #145, opened 2019-06-11
Add GA code

shammamah:
## About 

* Moved git user/email config out of the if conditional for R apps
* Used CircleCI environment variable to directly write to GA file in `assets/` folder 
* Changed CircleCI job to fail if there is no `assets/` folder in the app directory 
* Removed unnecessary code from the predeploy script (R apps now have the files copied over from within the Circle job and consequently don't need to be part of the predeploy script)
* Updated README to reflect that only Python apps can use the app name mapping file


predeploy.py
```diff
@@ -15,20 +14,7 @@
 if "DOKKU_SCALE" in os.listdir(app_path):
     shutil.copyfile(os.path.join(app_path, "DOKKU_SCALE"), "DOKKU_SCALE")
 
-
-file_list = rfiles if "-" in app_name and app_name.split("-")[0] == "dashr" else pyfiles
-
-for f in file_list:
+for f in pyfiles:
```
rpkyle: 👍 I think it's safe to :hocho: the R-related lines here, given that @tarzzz's fix uses the pre-predeploy.py CircleCI test flow to handle R app deployments.

README.md
```diff
@@ -46,7 +46,7 @@ There are two options when you are naming the folder:
 
 1. Make the folder have the _exact same_ name as the Dash app name.
 
-2. Select any other name, but _update the file
+2. (Python apps only) Select any other name, but _update the file
    [`apps_mapping.py`](apps_directory_mapping.py)_ with the Dash app
    name and the folder name you have selected.
```
rpkyle: I will open a second PR to update the README.md with the additional details that are currently required for R app deployments.

.circleci/config.yml
```diff
@@ -35,6 +35,7 @@ jobs:
           name: Deploy
           command: |
             APPS_MODIFIED=$(git diff origin/master origin/$CIRCLE_BRANCH --dirstat=files,1 apps/ | awk '{ split($2,a,"/"); if (length(a[2]) != 0) { print a[2]} } ' | sort -u)
+	    export PLOTLY_GA_CODE=$PLOTLY_GA_CODE
```
tarzzz: Please clean up the <tab> here.

.circleci/config.yml
```diff
@@ -35,6 +35,8 @@ jobs:
           name: Deploy
           command: |
             echo $PLOTLY_GA_CODE > ga_code.txt
+	    cat ga_code.txt
```
tarzzz: If it is confidential, it is not a good idea to `cat` it here. if it is not, I dont see a reason why it needs to go to circle CI environment variables.

predeploy.py
```diff
@@ -20,7 +20,7 @@
 
 for f in file_list:
     shutil.copyfile(os.path.join(app_path, f), f)
-    
+
```
tarzzz: Would be good to squash these commits into one because it is easier for the reviewer.

.circleci/config.yml
```diff
@@ -9,114 +9,115 @@ jobs:
     steps:
       - checkout
       - run:
-          name: install black
-          command: |
-            sudo pip install black
+	  name: install black
+	  command: |
+	    sudo pip install black
```
tarzzz: Here, you are actually converting spaces to tabs. Might want to fix that..

.circleci/config.yml
```diff
@@ -9,114 +9,115 @@ jobs:
...
-                  exit 0
-            fi
+	  name: Deploy
+	  command: |
+	    APPS_MODIFIED=$(git diff origin/master origin/$CIRCLE_BRANCH --dirstat=files,1 apps/ | awk '{ split($2,a,"/"); if (length(a[2]) != 0) { print a[2]} } ' | sort -u)
+	    if [ -z "$APPS_MODIFIED" ]
+	    then
+		  echo "No app change detected. Skipping the deploy.."
+		  exit 0
+	    fi
```
tarzzz: This whole commit seems to be doing opposite of what you mentioned in the commit message (ie, "change tabs to spaces"). You might want to remove this commit or clean it up.

shammamah:
@tarzzz Thank you! I'll make sure to do that in the future.

Issue #21, opened 2019-05-01
Lastodash [R]


shammamah:
# App name
dashr-lastodash 

# App
- [ ] This app is currently deployed on Dash(R) Gallery: [deployment URL]
- [x] This app is new

# Source code repository
[source code repo URL]

# Current appearance
[attach screenshot of current app here]

# Redesign
Link: https://github.com/plotly/streambed/issues/12068#issuecomment-469896383

![](https://user-images.githubusercontent.com/37411533/53983738-e8447600-40e5-11e9-9e4b-e360a05fa547.jpg)

# Playground deployment 
[url of playground deployment]

angeladaodao:
[LAS report.zip](https://github.com/plotly/dash-sample-apps/files/3183900/LAS.report.zip)

lundstrj:
Closing as per #400

PR #164, opened 2019-06-21
DashR LAStoDash

ohaanika:
Issue for app: https://github.com/plotly/dash-sample-apps/issues/21

# App pull request
- [x] This is a new app

## About

- Python app repository link: https://github.com/plotly/dash-sample-apps/tree/dash-lastodash/apps/dash-lastodash
- Python gallery app URL: https://dash-gallery.plotly.host/dash-lastodash/

## Workflow

  the style and functionality of the existing app.
  `assets/` folder.

## The pre-review review


apps/dashr-lastodash/app.R
```diff
@@ -0,0 +1,379 @@
...
+  A = "data.frame")
+)
+
+convert_las <- function(las_path) {
+  las_data <- las_object()
+  las_string <- read_file(las_path)
+  las_sections <- str_split(las_string, "~")
+  for (section in las_sections[[1]]) {
+    # store section containing version and wrap mode information
+    if (startsWith(section, "V")) {
```
rpkyle: @sacul-git I think this is why @ohaanika needs `gdata`; the package contains the `startsWith` function, which is Python-esque.

One way to rewrite this to eliminate the call to `startsWith` might be to try instead

```
> x <- "Test"
> if (substring(x, 1, 1) == "T") print("Found it!")
[1] "Found it!"
```

apps/dashr-lastodash/app.R
```diff
@@ -0,0 +1,379 @@
...
+          )
+        )
+      )
+    )
+  )
+}
+
+generate_table <- function() {
+  return(
+    dashTable::dashDataTable(
```
rpkyle: The namespace for `dashTable` should get loaded when you `library(dashTable)`, so we can :hocho: `dashTable::` here.

apps/dashr-lastodash/app.R
```diff
@@ -0,0 +1,379 @@
...
+        width = "25%",
+        textAlign = "center", 
+        border = "white"
+      ),
+      style_cell_conditional = list(list(
+        'if' = list(row_index = "even"), 
+        backgroundColor = "#f9f9f9"
+      )),
+      columns = unname(table_header),
+      data = dashTable::df_to_list(las_data$W[c("mnemonic", "description", "unit", "value")])
```
rpkyle: Same as above, can :hocho: `dashTable::`.

apps/dashr-lastodash/app.R
```diff
@@ -0,0 +1,379 @@
...
+    plots[[1]], plots[[2]], plots[[3]], plots[[4]], plots[[5]], 
+    get_plot_list(5), nrows = 1, shareY = TRUE, margin = 0
+  )
+  fig$x$data[[2]][["xaxis"]] = "x6"
+  fig$x$data[[7]][["xaxis"]] = "x7"
+  fig$x$data[[9]][["xaxis"]] = "x8"
+  fig$x$data[[12]][["xaxis"]] = "x9"
+  for (i in 1:length(xvals)) {
+    for (xval in xvals[[i]]) {
+      ifelse(i == 1, xaxis <- "xaxis", xaxis <- glue("xaxis{i}"))
```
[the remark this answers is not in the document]
rpkyle: Indeed, also, beware of lazy evaluation:

https://advanced-r-solutions.rbind.io/functions.html#lazy-evaluation
https://gist.github.com/jennybc/37481d9d784d2e8222b3

Assignments within function calls in R can have ... interesting results.

apps/dashr-lastodash/app.R
```diff
@@ -0,0 +1,379 @@
...
+)
+
+####################################################################################################
+# DEFINE APP CALLBACKS
+
+app$callback(
+  output = list(id = "las-table-print", property = "children"),
+  params = list(input(id = "table", property = "data")),
+  function (data) {
+    num_tables <- as.integer(length(data)/34)+1
```
rpkyle: @ohaanika I think I understand why it makes sense here, but I'm always nervous about fixed numeric values being used to subset by rows or columns. Whenever possible, I try to use character strings (either to match on rows or columns, then subset by the index returned from something like `match`), since hardcoded indices tend to be fragile.

I see `34` reappear in the `start` and `ifelse` blocks below too. Just pointing it out for the future.

apps/dashr-lastodash/app.R
```diff
@@ -0,0 +1,379 @@
...
+# IMPORT DATA
+
+las_dir <- "data/"
+las_file <- "alcor2.las"
+las_path <- paste(las_dir, las_file, sep = "")
+
+####################################################################################################
+# HANDLE & STORE DATA
+
+las_object <- setRefClass("las_object", fields = list(
```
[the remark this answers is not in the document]
ohaanika: Actually I'm converting back to a named list now. I was experimenting with `setRefClass` when I felt the need for more functions to convert the different sections of the LAS file which I planned to weave into the class but that's not necessary anymore!

apps/dashr-lastodash/app.R
sacul-git: That makes sense. `startsWith` was also added to [base in `3.3`](https://stat.ethz.ch/R-manual/R-devel/library/base/html/startsWith.html), not sure if the syntax is the same between `base` and `gdata` though.

ohaanika: Ah nice that's why it worked without `gdata`! Since they say `startsWith` is faster, I'll stick to `startsWith` but remove the `library(gdata)`.

apps/dashr-lastodash/app.R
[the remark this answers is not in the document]
ohaanika: Such a good point, correcting habits now!

apps/dashr-lastodash/app.R
```diff
@@ -0,0 +1,379 @@
...
+  # y-axis
+  fig$x$layout[["yaxis"]]$mirror <- "all"
+  fig$x$layout[["yaxis"]]$automargin <- TRUE
+  fig$x$layout[["yaxis"]]$showline <- TRUE
+  fig$x$layout[["yaxis"]]$title <- list(text = generate_axis_title(las_data$C, "DEPT"), font = list(family = "Arial, sans-serif", size = font_size))
+  fig$x$layout[["yaxis"]]$tickfont <- list(family = "Arial, sans-serif", size = tick_font_size)
+  fig$x$layout[["yaxis"]]$autorange <- 'reversed'
+  # layout
+  fig$x$layout$height <- height
+  fig$x$layout$width <- width
```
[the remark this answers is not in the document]
ohaanika: I had that because it would increase the width of the table in the printable pdf. Maybe I can just create a callback to allow resizing when the print button is clicked since I agree, a wider graph would be more preferable in the app itself.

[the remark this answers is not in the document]
rpkyle: Looks OK as is;  we can revisit if needed later, but the UI and features are 👌 for now!

ohaanika:
Other than renaming `filtering` and `sorting` to `filter_action` and `sort_action` respectively, do we have to add any additional properties for the sort and filter functions of the dash table to work?

PR #167, opened 2019-06-21
add pre-commit config for black

chriskfwoo:
Every time you commit, it will run the black checker.

1. install pre-commit (https://pre-commit.com/)
2. run `pre-commit install` to install the hooks in `.git/` folder

image

@shammamah you think this will be useful? If not, we can close this PR 😄

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,8 @@
+repos:
+-   repo: https://github.com/python/black
+    rev: stable
+    hooks:
+    - id: black
+      args: [--check]
+      exclude: 'venv/'
+      language_version: python3.6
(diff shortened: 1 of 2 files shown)


chriskfwoo:
> @chriskfwoo I think this is a great idea! However, let's not put the YAML file in the repo. Could you instead put in the README an example of a `pre-commit` config file? Or does it need to be part of the repo?

@shammamah The [pre-commit](https://pre-commit.com/) extension requires a YAML file. We could either put it in the README and it's up to the dev to create the file. 

I think the only benefit of having it in the repo is that it might force developers to use it more. And, the config for black (we can set different rules) will be the same across all developers (maybe more consistent?).

josegonzalez:
As per https://github.com/plotly/streambed/issues/13854, I am closing this as the new deployment process is in place.

If you need any changes from this branch, please recreate your changes against master and create a new pull request. Let me know if you need help and I am happy to jump in.

Issue #97, opened 2019-05-21
Animal Study Browser [PYTHON]


shammamah:
# App name
dash-animal-study-browser

# App
- [ ] This app is currently deployed on Dash(R) Gallery: [deployment URL]
- [x] This app is new

# Source code repository
https://github.com/shday/studybrowser

# Current appearance
![image](https://user-images.githubusercontent.com/15622051/58118999-05122500-7bd0-11e9-80e9-eeb78f4239f5.png)

# Redesign
![Animal Study Browser-01](https://user-images.githubusercontent.com/37411533/58222521-4f84c600-7ce4-11e9-81e4-4e9350d64e39.jpg)

# Playground deployment 
[url of playground deployment]

angeladaodao:
[Animal Study Browser.ai.zip](https://github.com/plotly/dash-sample-apps/files/3210081/Animal.Study.Browser.ai.zip)

shammamah:
@ronniebugia Please note that there have been some updates to the code here: https://github.com/shday/studybrowser

Can you integrate these changes into your app and then work on the redesign?

PR #170, opened 2019-06-25
Dash Animal Study Browser

ronniebugia:
Issue for app: #97 

# App pull request
- [ ] This is a new app

## About

- Playground deployment URL (new version): https://dash-playground.plotly.host/dash-study-browser

## Workflow

  `assets/` folder.

## The pre-review review


apps/dash-study-browser/app.py
```diff
@@ -0,0 +1,322 @@
...
+                        )
+                    ],
+                ),
+            ],
+        ),
+    ]
+)
+
+# Callback to create error message
+@app.callback(Output("error-message", "children"), [Input("upload-data", "contents")])
```
shammamah:
```suggestion
@app.callback(
    Output("error-message", "children"), 
    [Input("upload-data", "contents")])
```

Just for consistency with the other callback below!

apps/dash-study-browser/app.py
```diff
@@ -0,0 +1,322 @@
...
+            for study in studies:
+                options.append(
+                    {
+                        "label": "{} (study: {})".format(test_article, study),
+                        "value": study,
+                    }
+                )
+    else:
+        studies = study_data.study_id.unique()
+        for study in studies:
```
chriskfwoo: This is preference but if ur using the variable only once.

I prefer `for study in study_data.study_id.unique():`

apps/dash-study-browser/app.py
ronniebugia: I think that was formatted by `black`

apps/dash-study-browser/app.py
```diff
@@ -0,0 +1,322 @@
...
+            study_data.group_id[study_data.group_type == "reference"].unique()
+        )
+        control_groups = set(
+            study_data.group_id[study_data.group_type == "control"].unique()
+        )
+        all_groups = set(study_data.group_id.unique())
+        groups_to_annotate = all_groups - ref_groups - control_groups
+        annotations = [
+            dict(
+                x=test_stats[group_id]["index"],
```
chriskfwoo: if `study_data.group_id` is empty, it won't loop and add data to the `tests_stats`.

This will cause a index error `test_stats[group_id]["index"]`. I'm not sure if this edge case will appear often.

apps/dash-study-browser/app.py
```diff
@@ -0,0 +1,322 @@
...
+@app.callback(
+    [Output("study-dropdown", "options"), Output("study-dropdown", "value")],
+    [Input("error-message", "children")],
+    [State("upload-data", "contents")],
+)
+def update_dropdown(error_message, contents):
+    if error_message:
+        study_data = default_study_data
+    elif contents:
+        content_string = contents.split(",")
```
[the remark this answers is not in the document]
ronniebugia: good suggestion

apps/dash-study-browser/app.py
ronniebugia: I don't think this case will happen, because `group_id` is a necessary prereq for the study data

shammamah: It's a good idea to program defensively! You can use the `.get()` method for Python dictionaries  to return a default value if `group_id` isn't in `test_stats`.

apps/dash-study-browser/app.py
```diff
@@ -0,0 +1,326 @@
...
+    # Check if there is uploaded content
+    if contents:
+        content_type, content_string = contents.split(",")
+        decoded = base64.b64decode(content_string)
+
+        # Try reading uploaded file
+        try:
+            study_data = pd.read_csv(io.StringIO(decoded.decode("utf-8")))
+
+            # Missing columns
```
shammamah: This comment is unnecessary!

apps/dash-study-browser/app.py
```diff
@@ -0,0 +1,326 @@
...
+                                            },
+                                        ),
+                                    ],
+                                ),
+                                html.Div(
+                                    className="padding-top-bot",
+                                    children=[
+                                        html.H6("CSV File"),
+                                        dcc.Upload(
+                                            id="upload-data",
```
shammamah: You can specify filetypes that are accepted (still a good idea to keep that check in the callback, but this way you will avoid issues that stem from non-CSV files being uploaded). https://dash.plot.ly/dash-core-components/upload (`accept` argument)

apps/dash-study-browser/app.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+# Callback to generate study data
+@app.callback(
+    Output("plot", "figure"),
+    [Input("chart-type", "value"), Input("study-dropdown", "value")],
+    [State("upload-data", "contents"), State("error-message", "children")],
+)
+def update_output(chart_type, study, contents, error):
+    if error or not contents:
```
shammamah: Can you try putting the state of the error in a `dcc.Store` instead? This is kind of a weird pattern to use (checking to see if an HTML component's children is `None`).

apps/dash-study-browser/app.py
ronniebugia: I have a test in the error message callback to see if `group_id  in the headers is in the csv file. If not then use a default data set.
```
            missing_columns = {
                "group_id",
                "group_type",
                "reading_value",
                "study_id",
            }.difference(study_data.columns)
```

Also if the csv file has the headers, but is empty the `plot` figure callback should return an empty dictionary for the figure. 

But I added the `.get()` check anyway just to be sure.

chriskfwoo:
Please update your dash requirement to `dash==1.0.0`. 😄

PR #186, opened 2019-07-02
Fix dashr-wind-streaming app merge

rpkyle:
This is an attempt to redeploy `dashr-wind-streaming` from a fresh branch, as @shammamah and I continue to investigate why the earlier two attempts were not successful.


rpkyle:
I'm going to :dancer: this in the interests of time.

PR #203, opened 2019-07-04
Prevent dashr deploys

shammamah:


.circleci/config.yml
```diff
@@ -93,6 +87,10 @@ jobs:
             APPS_MODIFIED=$(git diff origin/production origin/master --dirstat=files,1 apps/ | awk '{ split($2,a,"/"); if (length(a[2]) != 0) { print a[2]} }' | sort -u)
             for APP in $APPS_MODIFIED
             do
+              if [[ $APP =~ "dashr" ]]
```
rpkyle: I had wondered if we should handle `dashR` or `DashR` as well, but part of me thinks either should be considered a CircleCI error that prevents deployment. Any thoughts?

shammamah: All of those converted to lowercase would just be "dashr", so we could do that instead before checking for this regex?

rpkyle: Yeah, I think that makes sense. My thought was that we could be somewhat heavy-handed and require that app developers rename their branches before the tests would proceed. But that seems somewhat ... extreme.

rpkyle:
LBTD! (Let's block those deployments.) 💃

PR #210, opened 2019-07-05
Dash molecule 3d viewer

z267xu:
Dash R molecule 3d viewer


shammamah:
Can you please add in an `assets/` folder and a link to the playground deployment in the initial PR comment? The missing folder is what's causing the deployment to fail.

PR #237, opened 2019-07-19
Dash yield curve

celinehuang:
Issue for app: #9 

# App pull request
- [ ] This is a new app

## About

- Playground deployment URL (new version): https://dash-playground.plotly.host/dash-yield-curve
- Current gallery app URL:  https://dash-gallery.plotly.host/dash-yield-curve

## Workflow

image

  `assets/` folder.

## The pre-review review


apps/dash-yield-curve/app.py
```diff
@@ -0,0 +1,425 @@
...
+def advance_slider(back, nxt, slider):
+
+    if back is None:
+        back = 0
+    if nxt is None:
+        nxt = 0
+    if slider is None:
+        slider = 0
+
+    global last_back
```
shammamah: Please use `dcc.Store` for these instead!

shammamah: This is leading to some odd behaviour; when you click on another point in the slider, then click back to the first point, then hit next, nothing happens.

PR #238, opened 2019-07-19
Dashr molecule 3d viewer

z267xu:
Remove browser from `app.R`


rpkyle:
> @rpkyle Might have more insight as to what removing `browser()` does to the app, but looks good to me :)

Definitely something I should add to CircleCI, since this is a problem I've had a few times too many myself 😂 

@shammamah Zehao is removing a line which invoked an interactive debugger from his app, so looks 👌 to me 🙂

PR #240, opened 2019-07-19
DashR Ocean Optics

CanerIrfanoglu:
# App pull request
- [x] This is a new app

## About

- Current gallery app URL: https://dash-gallery.plotly.host/dash-ocean-optics/
- Python app repository link: https://github.com/plotly/dash-ocean-optics

## Workflow

  `assets/` folder.

## The pre-review review


apps/dashr-ocean-optics/app.R
```diff
@@ -0,0 +1,580 @@
...
+             list(state(id = "power-button", "on"))),
+
+  function(...) {
+    args <- list(...)
+
+    # handle `NULL` args`
+    nullCheck <- sapply(args, class)
+    nullCheck == "list"
+    if (sum(nullCheck == "list") > 0 ) {
+      args[[which(nullCheck == "list")]] <- "None"
```
rpkyle: Here, I'd expect that we want to assign `NULL` instead of `"None"`, as the latter is a Python sort of representation for null objects.

The line `nullCheck == "list"` looks inert; I think this is just comparing whether `nullCheck` is identical to `"list"`, but nothing is done with this information. Probably safe to :hocho:.

For `sum(nullCheck == "list") > 0`, I might actually rewrite this as `if ("list" %in% nullCheck)`. The reason for this is that the latter will work regardless of the dimensionality of `nullCheck`:

```
> nullCheck <- c("list", "character", "numeric")
> "list" %in% nullCheck
[1] TRUE
> nullCheck <- NA
> "list" %in% nullCheck
[1] FALSE
> sum(nullCheck)
[1] NA
> sum(nullCheck) > 0
[1] NA
```

apps/dashr-ocean-optics/app.R
```diff
@@ -0,0 +1,580 @@
...
+      }
+
+      sumList <-  c(sumList,
+                    list("The following parameters were successfully updated:"),
+                    nextLine)
+
+      # append updated controls to sumList
+      for (i in 1:length(controls)) {
+
+        if (args[[i + 1]] == "l1") {
```
rpkyle: Instead of using `next` to modify control flow here, I would probably handle this by subsetting `controls` to operate only on the elements for which `args[[i + 1]] == "l1"` is `FALSE`.

apps/dashr-ocean-optics/app.R
```diff
@@ -0,0 +1,580 @@
...
+    input(id = "submit-button", "n_clicks_timestamp"),
+    state(id = "light-source-input", property = "value"),
+    state(id = "power-button", property = "on")
+  ),
+
+  function(clicks, value, on) {
+    if (on) {
+      # Handle the case when value is removed
+      if ( is.list(value) ) {
+        list("None")
```
rpkyle: See comment above about `"None"` versus `NULL`.

apps/dashr-ocean-optics/app.R
```diff
@@ -0,0 +1,580 @@
...
+    state(id = "power-button", property = "on"),
+    state(id = "autoscale-switch", property = "on"),
+    state(id = "light-intensity-knob", property = "value"),
+    state(id = "hidden-div-int-time", property = "children"),
+    state(id = "hidden-div-light-source", property = "children")
+  ),
+
+  function(n, pwr, auto_range, knob_intensity, scale, ls) {
+
+    ls <- unlist(ls)
```
rpkyle: Rename `ls`, since it's a base R function name.

apps/dashr-ocean-optics/app.R
```diff
@@ -0,0 +1,580 @@
...
+      layout = layout)
+    )
+  }
+)
+
+if (!appName == ""){
+  app$run_server(host = "0.0.0.0", port = Sys.getenv('PORT', 8050))
+} else {
+  app$run_server(debug = TRUE)
+}
```
rpkyle: If you're using RStudio, would you mind enabling newlines at EOF in the preferences? It will make the visual warning in GitHub that currently appears go away 🙂  I'd tick the second box as well, since it will save you from linting errors if you edit Python in RStudio too.

Screenshot 2019-07-22 19 10 29

If you're using another editor, there's likely to be a similar setting there too.

apps/dashr-ocean-optics/init.R
```diff
@@ -0,0 +1,41 @@
...
+install.packages("https://cloud.r-project.org/src/contrib/codetools_0.2-16.tar.gz", type="source", repos=NULL)
+install.packages("https://cloud.r-project.org/src/contrib/globals_0.12.4.tar.gz", type="source", repos=NULL)
+install.packages("https://cloud.r-project.org/src/contrib/Archive/future/future_1.11.1.1.tar.gz", type="source", repos=NULL)
+# fiery and friends
+install.packages("https://cloud.r-project.org/src/contrib/routr_0.3.0.tar.gz", type="source", repos=NULL)
+install.packages("https://cloud.r-project.org/src/contrib/fiery_1.1.1.tar.gz", type="source", repos=NULL)
+
+remotes::install_github("plotly/dash-html-components")
+remotes::install_github("plotly/dash-core-components")
+remotes::install_github("plotly/dash-daq")
```
rpkyle: Might be safer to swap the order of `dash-daq` and `dashR` for now, I'm not certain that `dash-daq` will auto-install the package from `dashR` quite yet.

apps/dashr-ocean-optics/app.R
```diff
@@ -0,0 +1,580 @@
...
+
+  setwd(sprintf("/app/apps/%s", appName))
+}
+
+# colors for layout and figure
+colorsRaw <- read.table("colors.txt", sep = ' ', comment.char = '',
+  stringsAsFactors = FALSE)
+
+colors <- list()
+colors <- sapply(1:nrow(colorsRaw), function(i) {
```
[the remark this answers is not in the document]
CanerIrfanoglu: Thanks. Replaced with `setNames` version. Looks much more compact indeed.

apps/dashr-ocean-optics/app.R
CanerIrfanoglu: Assigning directly `NULL` to `args[[which(nullCheck == "list")]] <- NULL` results in corresponding `args` index to be removed from the list. Since it causes in the following `for` loop assignments to shift, I replaced `"None"` to `"NULL"` to follow R representation.

https://github.com/plotly/dash-sample-apps/blob/c04df66e9187b950f4dd850932b88580cd2aa8c8/apps/dashr-ocean-optics/app.R#L419-L435

apps/dashr-ocean-optics/app.R
```diff
@@ -0,0 +1,580 @@
...
+# disable/enable intensity knob
+app$callback(
+  output = list(id = "light-intensity-knob", property = "disabled"),
+  params = list(
+    input(id = "power-button", property = "on"),
+    input(id = "light-source-input", property = "value")
+  ),
+
+  function(pwr, ls) {
+    return(!(pwr && ls != "NULL" && ls != ""))
```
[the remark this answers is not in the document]
CanerIrfanoglu: Changed to:
```
  function(pwr, lsi) {
    return(!(pwr && !is.null(lsi[[1]]) && lsi != ""))
  }
)
```
Since `lsi` is type `list` when the `dccDropdown` is empty and `character` otherwise, had to change `lsi` to `lsi[[1]]`.

CanerIrfanoglu:
@rpkyle thanks for the review. I applied all the changes suggested previously in https://github.com/plotly/dash-sample-apps/pull/240/commits/b6d5b9e389584797bd4d63f2f5b20dac045dc3b9.

apps/dashr-ocean-optics/app.R
```diff
@@ -0,0 +1,574 @@
...
+  pathPrefix <- sprintf("/%s/", appName)
+
+  Sys.setenv(DASH_ROUTES_PATHNAME_PREFIX = pathPrefix,
+             DASH_REQUESTS_PATHNAME_PREFIX = pathPrefix)
+
+  setwd(sprintf("/app/apps/%s", appName))
+}
+
+# colors for layout and figure
+colorsRaw <- read.table("colors.txt", sep = ' ', comment.char = '',
```
sacul-git: Not an important comment, but just out of curiosity, is there a reason to store the color data in a separate file? Why not just use a list like:

```
colors <- list(
  background = "#bbbbbb",
  primary = "#ffffff",
  secondary = "#efefef",
  tertiary = "#dfdfdf",
  grid-colour = "#eeeeee",
  accent = "#2222ff"
)
```

CanerIrfanoglu: Since Python version was using a separate `colors.txt`, I did it the same way. Also, I wouldn't mind defining it as a `list` in app.R :)

Issue #61, opened 2019-05-02
MapD Example [PYTHON]


shammamah:
# App name
dash-mapd-demo

# App
- [ ] This app is currently deployed on Dash(R) Gallery: [deployment URL]
- [x] This app is new

# Source code repository
https://github.com/plotly/dash-sample-apps/tree/dash-mapd-demo

# Current appearance
<img width="1424" alt="Screen Shot 2019-05-15 at 1 33 45 PM" src="https://user-images.githubusercontent.com/35306149/57796376-2c6f7a80-7716-11e9-838b-46d5d8229100.png">

# Redesign
![US Flight-01-min](https://user-images.githubusercontent.com/37411533/58284699-29f5cc00-7d79-11e9-855a-9c057c8cf184.jpg)

# Playground deployment 
[url of playground deployment]

angeladaodao:
[US Flight.ai.zip](https://github.com/plotly/dash-sample-apps/files/3214074/US.Flight.ai.zip)

PR #244, opened 2019-07-23
update to dash 1.0.2 (dash-mapd-demo)

celinehuang:
Issue for app: #61 
- update requirements.txt to latest dependencies
- Remove ._dash-undo-redo css

https://dash-playground.plotly.host/dash-mapd-demo/


ycaokris:
@celinehuang  I've removed DASH_APP_NAME so that it's not being added as url route prefix locally. To run this app locally, set environment variable DB_HOST, and `python app.py`. 

Looks like loading on playground is good now, I'll open another PR for performance investigation. This could be merged for now.

PR #247, opened 2019-07-26
Fix table cell white border artifact

ycaokris:
## About

In dash-table==0.4.2,  background color of table-row is set to be white, and overwrites border color at different zooming ratio in presence of css rule as {"selector": "td", "rule": "border: none !important;"}. 

datatable-border

```
spreadsheet-inner tr {
    background-color: white; 
}
```

Fixed by setting it transparent.


ycaokris:
Deployment fails caused by server downtime. Similar broken task https://circleci.com/gh/plotly/dash-sample-apps/4120
Screen Shot 2019-07-26 at 5 43 32 PM

ycaokris:
Fix in https://github.com/plotly/dash-table/pull/521

PR #265, opened 2019-08-02
Update Dash logos in all dashBio apps

HammadTheOne:
Issue - Dash Bio app logo's have been updated to reflect the new Plotly - Dash logo in the python version. This PR fixes the headers and updates the logo for all Dash-Bio R app versions and adds the logo to their corresponding assets.

# App pull request
- [ ] This is a new app

## Workflow

  `assets/` folder.

## The pre-review review


apps/dashr-clustergram/app.R
```diff
@@ -41,13 +41,12 @@ header <- htmlDiv(
...
-        htmlImg(
-          src = "assets/dashbio_logo_transparent.png" 
-        )
+        htmlImg(src='assets/plotly-dash-bio-logo.png', height = '36', width = '180',
+                style = list('top' = '10', 'margin-left' = '10px'))
       ),
       href = "/Portal"
     ),
-    htmlH2("Dash Clustergram"),
+    htmlH2("    Clustergram"),
```
shammamah: Why is there still whitespace here?

apps/dashr-financial-report/app.R
```diff
@@ -782,7 +782,7 @@ dividends_data[2:6,8] <- "-"
 
 
 
-dividends_table <- generate_table(remove.factors(dividends_data))
+dividends_table <- generate_table(dividends_data)
```
shammamah: Not sure why this diff is here -- this PR looks like it's for the Dash Bio apps. Might be a merge issue with `master`?

apps/dashr-clustergram/app.R
HammadTheOne: I left in the whitespace to act as some spacing from the logo so it's not side-by-side. Another way to achieve the same thing would be to add padding to either container. Would that be a better alternative?

shammamah: Ah I see! Yes, in this case we should be using CSS. Otherwise, it isn't super clear what you're trying to do.

apps/dashr-financial-report/app.R
HammadTheOne: I was just going through all apps I'd worked on to check the logo used on them and noticed this small bug. There's no real changes with the app itself, but should I open a separate PR for this?

apps/dashr-clustergram/app.R
HammadTheOne: Fixed!

apps/dashr-financial-report/app.R
rpkyle: @HammadTheOne Is it possible to ignore this file and remove the directory from your commit (without erasing the existing app, of course)? Even if it triggers a conflict, we can resolve it before merging.

HammadTheOne: Should be fixed now, I reverted the change from this commit. Is that OK?

PR #273, opened 2019-08-07
Dash svm fix demo link

YunkeXiao:


shammamah:
@YunkeXiao Could you actually make these changes on a separate branch (that isn't one of the app branches)? Ideally the `dash-svm` branch should contain changes to *only* `dash-svm`. Other than that this looks good :)

PR #275, opened 2019-08-07
DashR DAQ IV Tracer

CanerIrfanoglu:

# App pull request
- [x] This is a new app

## About

- Current gallery app URL: https://dash-gallery.plotly.host/dash-daq-iv-tracer/
- Python app repository link: https://github.com/plotly/dash-sample-apps/tree/master/apps/dash-daq-iv-tracer

## Workflow

  `assets/` folder.

## The pre-review review


apps/dashr-daq-iv-tracer/app.R
```diff
@@ -0,0 +1,1508 @@
...
+        plot_bgcolor = cardColor[[theme]],
+        paper_bgcolor = cardColor[[theme]]
+      )
+    )
+
+    if (!(modeChoice)) { # Single measure case
+      if (measTriggered) {
+
+        # Single data point does not draw graph,
+        # duplicating data to show single data point
```
rpkyle: Interesting; why is this step necessary?

CanerIrfanoglu: When the voltage/current value set from the `daqKnob` and `Single Measure` is clicked, we expect having a single data point on the graph like in the python version. However, when there's only 1 data point in `xdata` and `ydata`, the callback was returning an empty graph. Just duplicating values in this scenario solved the issue. However, not sure if it is the best solution.

Screen Shot 2019-08-08 at 2 55 07 PM

CanerIrfanoglu:
@rpkyle thank you for reviewing. Just pushed commits https://github.com/plotly/dash-sample-apps/pull/275/commits/1e3e188e2c4ab12bb2dbc0fbdee0c4392ca63124 https://github.com/plotly/dash-sample-apps/pull/275/commits/d686966f94392c10f320b774632859eb5f762e74 applying the suggested changes.

apps/dashr-daq-iv-tracer/app.R
rpkyle: Ah, OK; thanks for the additional detail. I think we should revisit this sometime soon, but for now it's fine. We can discuss alternatives and then just open a new, tiny PR to update later.

rpkyle:
Nice work!! 💃

CanerIrfanoglu:
@KPhans thank you for reviewing :) Solid suggestions, definitely will look into  `callback_context`. Guess `DashR-oil-and-gas` would be a good example.

Issue #337, opened 2019-10-17
dash-oil-and-gas map doesn't load


cldougl:
# App name
dash-oil-and-gas
# Description of bug
dash-oil-and-gas map won't load 
# How to replicate the bug
Open https://dash-gallery.plotly.host/dash-oil-and-gas/ (@sudburyrob saw this in his browser, I can only reproduce in incognito)
# Other notes

<img width="1278" alt="Screen Shot 2019-10-17 at 9 29 56 AM" src="https://user-images.githubusercontent.com/2789078/67014358-b905d680-f0c2-11e9-903c-a31e329ce9f3.png">

PR #338, opened 2019-10-17
Dash oil and gas app fix

ycaokris:
fixes https://github.com/plotly/dash-sample-apps/issues/337

Updated deployment on playground **https://dash-playground.plotly.host/dash-oil-and-gas/**

###### Problem:
1. map doesn't load because of mishandling of `relayoutData` inside callback - {'autosize': True} is not considered and key is not correct.

Screen Shot 2019-10-17 at 11 04 32 AM

After fix mapbox should be loading upon app loading,  and when controls in control panel is updated.

2. Fix a math error raised when `Customized` is set in well-status and no item/category in dropdown.

### DIFF
--- a/apps/dash-oil-and-gas/app.py
+++ b/apps/dash-oil-and-gas/app.py
@@ -162,6 +162,7 @@ app.layout = html.Div(
                             id="lock_selector",
                             options=[{"label": "Lock camera", "value": "locked"}],
                             className="dcc_control",
+                            value=[],
                         ),
                         html.P("Filter by well type:", className="control_label"),
                         dcc.RadioItems(
@@ -260,6 +261,8 @@ app.layout = html.Div(
 
 # Helper functions
 def human_format(num):
+    if num == 0:
+        return "0"
 
     magnitude = int(math.log(num, 1000))
     mantissa = str(int(num / (1000 ** magnitude)))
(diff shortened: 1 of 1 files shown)


apps/dash-oil-and-gas/app.py
```diff
@@ -450,14 +453,17 @@ def make_main_figure(
...
-        lon = float(main_graph_layout["mapbox"]["center"]["lon"])
-        lat = float(main_graph_layout["mapbox"]["center"]["lat"])
-        zoom = float(main_graph_layout["mapbox"]["zoom"])
-        layout["mapbox"]["center"]["lon"] = lon
-        layout["mapbox"]["center"]["lat"] = lat
-        layout["mapbox"]["zoom"] = zoom
+    if main_graph_layout is not None and selector is not None and "locked" in selector:
+        if (
+            "mapbox.center" in main_graph_layout.keys()
+        ):  # relayoutData is None by default, and {'autosize': True} without relayout action
```
cldougl: @ycaokris I think this might be a tiny bit easier to read if you format the comment above the if statement and make that 1 line:  
```
# relayoutData is None by default, and {'autosize': True} without relayout action
if "mapbox.center" in main_graph_layout.keys():
```
or incorporate it into the above if statement

Issue #333, opened 2019-10-09
dash-mapd-demo application error


cldougl:
# App name
dash-mapd-demo
# Description of bug
https://dash-gallery.plotly.host/dash-mapd-demo/ application error
More information in application logs: https://dash-gallery.plotly.host/Manager/apps/dash-mapd-demo/logs
# How to replicate the bug
Visit https://dash-gallery.plotly.host/dash-mapd-demo/
# Other notes

cldougl:
Suggested resolution:
1. add a cached fallback 
2. check if anything happened to the dB recently 
3. add retry logic around the connection

tarzzz:
> check if anything happened to the dB recently

I checked some logs, and it seems every time we try to access the DB, it crashes the container leading to container restart. I suspect this could be because this DB has `7009728` records (7M+) and not enough resources to handle that level of data.

I would recommend coming up with a smaller dataset (or use postgres/sqlite if that works) for this app.

cldougl:
> I suspect this could be because this DB has 7009728 records (7M+) and not enough resources to handle that level of data
@tarzzz why not allot more resources? (ie is this a pricing issue)
This app is supposed to demo the Omnisci DB connection so I don't think it quite resolves the problem to use postgres/sqlite

tarzzz:
> This app is supposed to demo the Omnisci DB connection so I don't think it quite resolves the problem to use postgres/sqlite

👍 . The top-comment ("add a cached fallback") suggested using postgres/sqlite so I thought that would be ok.

> why not allot more resources? (ie is this a pricing issue)

The records are loaded to the DB on container creation, so increasing the number of containers will not help to decrease the load. Currently each running container in the cluster has 7M records stored in it, which is very inefficient.

IMHO it would be better to use a smaller dataset (easier to maintain and better performance-wise) unless the app requires to query (or use) the 7M records for its functioning..

josegonzalez:
@tarzzz is the omnisci database still running, or is it crashing?

bpostlethwaite:
Any chance we can get an update on the status of this one? If we don't have time to fix it properly I think @cldougl is right, let's just pay more and create an issue to cut costs later.

tarzzz:
@bpostlethwaite The PR to use the smaller dataset is ready to be merged: https://github.com/plotly/dash-sample-apps/pull/339/

I am waiting on a 💃 on that one.

> is the omnisci database still running, or is it crashing?

[...]

PR #339, opened 2019-10-18
Use smaller dataset

tarzzz:
Issue for app: #333 

# App pull request

Specifying `2` in the prompt allows us to download the 10K dataset instead of the default 7M dataset.

### DIFF
--- a/apps/dash-mapd-demo/app.py
+++ b/apps/dash-mapd-demo/app.py
@@ -32,7 +32,7 @@ if "DB_HOST" in os.environ:
 else:
     host = "localhost"
 
-table = "flights_2008_7M"
+table = "flights_2008_10K"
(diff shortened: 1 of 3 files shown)


tarzzz:
@cldougl Please review.. As discussed, I have updated the omnisci deployment to use "10K" dataset instead of "7M". 

I have also increased resources on cluster (RAM increased from 4GB to 16GB) which should help with the performance.

tarzzz:
cc: @ycaokris .. !!

cldougl:
@tarzzz I've approved the pr already but 💃 if you were waiting

tarzzz:
I was waiting for Kris actually:
image

but it seems that is not required.. Merging..

PR #340, opened 2019-10-21
Dash nlp

ycaokris:

# App pull request
- [ ] This is a new app


apps/dash-nlp/app.py
```diff
@@ -272,17 +376,27 @@ def plotly_wordcloud(df):
         ],
         "layout": {"height": "550", "margin": dict(t=20, b=20, l=100, r=20, pad=4)},
     }
-
-    return wordcloud_figure_data, frequency_figure_data
+    treemap_trace = go.Treemap(
+        labels=word_list_top,
+        parents=[""] * len(word_list_top),
+        values=list(map(lambda x: x * 10, freq_list_top)),
```
ycaokris: @lundstrj  this is great 👍 .  Did you tried making this treemap without 10x?  The hover value on treemap is not consistent with those in wordcloud or in bar chart so it may be more clear to let them display same value. You may also try play with https://plot.ly/python/reference/#layout-yaxis-hoverformat in replacement of changing your values here.

ycaokris: Let me know if it works smoothly then we are good to merge this

lundstrj: I suppose it was a bit of a shortcut to use freq_list when I actually want the count of occurrences in the dataset (for that bank). Now when I tested it, I suppose the * 10 doesn't really add any value.

I'll remove it and consider using a different input.

Will look at the hoover format.

apps/dash-nlp/app.py
```diff
@@ -17,55 +17,113 @@
...
+    "Trans Union",
+    "BOA",
+    "Citi",
+    "account",
+]
+for stopword in additional_stopwords:
+    STOPWORDS.add(stopword)
 
+"""
+Produly written for Plotly by Vildly in 2019. info@vild.ly
```
rpkyle: @ycaokris Would change "Produly" to "Proudly" here. 🙂

apps/dash-nlp/app.py
```diff
@@ -17,55 +17,113 @@
...
+Users can at this point do deeper inspections into interesting formations or
+clusters in the scatter plot by zooming and clicking dots.
+
+Clicking on dots in the scatter plot will display a table showing the contents
+of the selected complaint (each dot is a specific complaint).
+
+It is worth mentioning that there is also a time frame selection slider which
+allows the user to look at specific time windows if there is desire to do so.
+
+To illustrate the usefullness of this dashboard we suggest looking at how the
```
rpkyle: @ycaokris Would change "usefullness" to "usefulness" here.

apps/dash-nlp/app.py
```diff
@@ -17,55 +17,113 @@
...
+wordcloud and scatter plot changes from Equifax if 2017 is included in the plots
+or not.
+
+Another potentially interesting find is that Capital One has a common word
+other banks seem to lack, "Macy". It would appear that Capital One at some point
+teamed up with popular retailer Macy's to offer their services. This campaing
+might have been hugely popular and thus explaining it's high frequency of occurance
+in complaints, or perhaps there are other reasons explaining the data.
+
+Rergardless of what caused these two mentioned outliers, it shows how a tool
```
rpkyle: @ycaokris Similarly, would replace "Rergardless" to "Regardless".

apps/dash-nlp/Procfile
```diff
@@ -1 +1 @@
-web: gunicorn --pythonpath apps/dash-nlp app:server
\ No newline at end of file
+web: gunicorn --pythonpath apps/dash-nlp app:SERVER
```
rpkyle: @ycaokris Super minor detail, but missing newline at EOF here.

apps/dash-nlp/app.py
```diff
@@ -727,6 +727,26 @@ def update_lda_table(value_drop, time_values, n_selection):
...
+    min_date = GLOBAL_DF["Date received"].min()
+    max_date = GLOBAL_DF["Date received"].max()
+    marks = make_marks_time_slider(min_date, max_date)
+    min_epoch = list(marks.keys())[0]
+    max_epoch = list(marks.keys())[-1]
+    bank_names, counts = get_complaint_count_by_company(GLOBAL_DF)
+    results = {}
+    time_values = [min_epoch, max_epoch]
+    n_selection = 100
+    file = open("precomupted", "w")
```
rpkyle: @ycaokris Should probably be "precomputed" here and below.

PR #344, opened 2019-10-29
DashR Uber Rasterizer

HammadTheOne:

# App pull request
- [x] This is a new app

## About

- Playground deployment URL (new version): https://dash-gallery.plotly.host/dashr-uber-rasterizer/
- Python app repository link: https://github.com/plotly/dash-sample-apps/apps/tree/dash-uber-rasterizer/

## Workflow

  `assets/` folder.

## The pre-review review


apps/dashr-uber-rasterizer/init.R
```diff
@@ -0,0 +1,13 @@
...
+# NOTE: this script is executed in the chroot context; check paths!
+
+r <- getOption("repos")
+r["CRAN"] <- "http://cloud.r-project.org"
+options(repos=r)
+install.packages("viridis")
+
+# ======================================================================
+# packages go here
+remotes::install_github("plotly/dashR", dependencies=TRUE)
```
rpkyle: @HammadTheOne I think we should use `upgrade=TRUE` instead.

apps/dashr-uber-rasterizer/app.R
```diff
@@ -0,0 +1,317 @@
...
+                            constrain = "domain"))
+    )
+  }
+)
+
+
+if(appName != "") {
+  app$run_server(host = "0.0.0.0", port = Sys.getenv('PORT', 8050))
+} else {
+  app$run_server(host = "127.0.0.1",
```
rpkyle: @HammadTheOne Here I would probably default to using `debug=TRUE` and setting the host and port, but maybe not enabling the other options. `viewer = TRUE` has changed since hot reloading support was initially available, and is now `use_viewer = TRUE`, but this feature won't be available until `dev` gets merged into `master`.

HammadTheOne: Agreed, fixed.

PR #354, opened 2019-11-12
[sample-apps] Dash-NLP: Precomputed LDA file and fix the app

oxananu:
I am improving an existing app Dash-NLP

### DIFF
--- a/apps/dash-nlp/app.py
+++ b/apps/dash-nlp/app.py
@@ -4,6 +4,7 @@ Module doc string
 """
 import pathlib
 import re
+import json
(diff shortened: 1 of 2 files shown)


lundstrj:
I can't merge this as it is at the moment.

- [ ] It throws an error
- [ ] The LDA scatter plot does not honor the time frame selection

> dash.exceptions.InvalidCallbackReturnValue: Invalid number of output values for ..lda-table.data...lda-table.columns...tsne-lda.figure...no-data-alert-lda.style...
>  Expected 4 got 3

PR #355, opened 2019-11-12
New google tags

ycaokris:


ycaokris:
Close this since changes needed for adding google tags have already been merged by #351.

PR #356, opened 2019-11-12
#351 follow up

ycaokris:
https://github.com/plotly/dash-sample-apps/pull/351 made a test deploy of adding new google tag upon playground deployment. 

I went through list in above deployment, and found out that I've missed initializing part of app list (my bad) on server panel.
Screen Shot 2019-11-12 at 5 02 28 PM

For apps that have been deployed successfully, only one app `dash-nlp` is encountering problem because it declared `SERVER` instead of `server` in Procfile so `predeploy` is unable to locate `app:SERVER`. Handled by https://github.com/plotly/dash-sample-apps/commit/17cd6c840c71178e95138d6fbf552cd50ce8eeb8

This PR does minor fix including:
- [x] Add missing python apps for a re-deploy validation
- [x] Remove unnecessary old js git add command
- [x] We missed a condition of apps carrying enviromnental variable as app settings.  `dash-salesforce-crm` and `dash-image-processing` are two examples I could think about, and I've added for both on test server app settings.

 @shammamah  Let me know whether we could get this merged before meeting with @VeraZab for a checkup.

### DIFF
--- a/predeploy.py
+++ b/predeploy.py
@@ -44,8 +44,10 @@ app_file_name = ""
 with open(os.path.join(app_path, "Procfile"), "r") as f:
     contents = f.read().split(" ")
     for item in contents:
-        if "server" in item:
+        if "server" or "SERVER" in item:
             app_file_name = item.split(":")[0]
+        else:
+            print("'server' or 'SERVER' not found in Procfile")
 
 full_app_path = os.path.join(app_path, app_file_name + ".py")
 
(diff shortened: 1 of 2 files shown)


predeploy.py
```diff
@@ -44,8 +44,10 @@
 with open(os.path.join(app_path, "Procfile"), "r") as f:
     contents = f.read().split(" ")
     for item in contents:
-        if "server" in item:
+        if "server" or "SERVER" in item:
```
shammamah: We should just edit the Procfile for that one app, since it's the only app for which this is the case.

ycaokris: > We should just edit the Procfile for that one app, since it's the only app for which this is the case.

This works as well. I could drop this edit and edit nlp Procfile.

PR #388, opened 2020-01-15
Add flake8 to CircleCI and enable rule E711.

jbampton:
Comparison to none should be 'if cond is none:'.

Lint Python code for rule E711.

https://www.flake8rules.com/rules/E711.html

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,13 +9,17 @@ jobs:
     steps:
       - checkout
       - run:
-          name: install black
+          name: install black and flake8
           command: |
-            sudo pip install black
+            sudo pip install black flake8
       - run:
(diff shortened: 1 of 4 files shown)


lundstrj:
This looks like a straight forward enough PR. We should be able to take care of it relatively shortly.

lundstrj:
@josegonzalez Hey man, what do you think about this one?

josegonzalez:
Seems reasonable to me. Any reason why we wouldn't want to apply _all_ of the flake8 rules, @jbampton ? Other than maybe the number of code changes (which is reasonable).

lundstrj:
Good enough for @josegonzalez is good enough for me.

PR #421, opened 2020-04-13
dash-nlp in R version

merenlin:
This is an R version of the dash-nlp app. 

Not sure that deployment will work but want to start it and get going from here. 

Let me know if we need to rework something.

# App pull request
- [ X] This is a new app


xhlulu:
We are currently cleaning this repository using BFG, which requires all open PRs to be closed. Feel free to recreate a new PR with the same commits after the repository is cleaned.

PR #428, opened 2020-05-04
Update api.py to functioning SQL queries

Xaimerr:
# App pull request
- [ ] This is a new app
- [X ] I am improving an existing app (redesigns/code "makeovers")

## The pre-review review
SQL query didn't work, just needed a minor change. App works great otherwise.

### DIFF
--- a/apps/dash-wind-streaming/db/api.py
+++ b/apps/dash-wind-streaming/db/api.py
@@ -2,7 +2,6 @@ import pathlib
 import sqlite3
 import pandas as pd
 
-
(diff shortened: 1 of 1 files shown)


josegonzalez:
> Note that unless the app has a specific `runtime.txt`, at this time, we test all apps with 3.6.2, as shown [here](https://github.com/plotly/dash-sample-apps/blob/master/runtime.txt).

I believe the app here is meant to work with Python 3.6+, which supports [f-strings](https://realpython.com/python-f-strings/). If at all possible, I would recommend switching to a supported Python release, as [Python 2 is EOL](https://pythonclock.org/).

PR #448, opened 2020-06-09
ct exploration app

emmanuelle:
# App pull request
- [x] This is a new app

## About

## Workflow

- [ ] I have created a branch in the appropriate monorepo, and the
  `assets/` folder.

## The pre-review review

- [ ] Does everything in my code serve some purpose? (I have removed
- [ ] Does everything in my code have a clear purpose? (My code is
- [ ] Am I reinventing the wheel? (I have used appropriate packages to


emmanuelle:
I will need a deployment of this app for the webinar but this is not ready yet to be reviewed.

emmanuelle:
@nicholas-esterer I finally finished a version of the app with which I'm happy :-). Please review when you have the time; I'm interested in knowing whether it's clear what to do from the different titles of figures, or whether more explanations should be added.

nicholas-esterer:
(Sorry I accidentally posted this in the wrong place before):
Hmm I'm trying to review the PR and after making the venv and installing the requirements, running with `python app.py` or `gunicorn app:server`, the app looks like this:
Screenshot from 2020-09-24 12-28-13
Maybe something is missing? :thinking:

emmanuelle:
Hum, thanks for reporting this. On playground it looks like this
image

Could you please try with the most recent stable version of Dash? Maybe I need to pin its version because it will only work with the most recent stable Dash.

nicholas-esterer:
Yes that fixed it. I just ran `pip install -U dash`. The specific version is `dash==1.16.1`.

nicholas-esterer:
As for the use of the app, it's pretty slick! I like that you can experiment with different selections of the histogram values, to see how that changes the selection.
A couple things that might be a little unclear to new users:
- the shapes drawn on the images stay when you scroll to other slices of the image, is that because they extend the full length of the image in the direction normal (in the linear algebra sense) to the shape? If so maybe some clarification is needed?
- I'm not sure if I can go back and edit the shapes and have the final result updated. For example if I try and erase the shapes after having chosen histogram values, they don't seem to erase... If users really need to do the process in order, maybe steps could be "greyed" out before one is allowed to proceed to them?

emmanuelle:
Thanks for the review Nick! Yes, annotations persist because the volume considered for the histogram segmentation is the "extrusion" of the closed path drawn on the left image, between heights defined by the rectangle drawn on the right image. This is a bit tricky to explain but I'll try to find a way to make this clearer. Also I agree that going back to correct annotations does not work well, I'll look into this.

emmanuelle:
@nicholas-esterer I have made the following changes
- it is now possible to modify a shape (one has to re-select a range in the histogram)
- I've added tooltips (with `dbc.Tooltip`) to add more explanations about the shapes to draw

nicholas-esterer:
Cool! It seems there is a small bug when drawing a rectangle in the 2nd figure from bottom to top: in that case the 'y0' and 'y1' points of the rectangle are switched. But I have a two-line fix to this problem (just sort the values before choosing top and bottom), should I just push it to this PR?

emmanuelle:
@nicholas-esterer thanks for noticing this bug! Yes please push a commit to this PR, thanks a lot!

nicholas-esterer:
I think I fixed it for when the users input funny shapes (upside down rectangles or shapes that are too small so that the slice indices for the start and end are the same).
I thought it might be nice to be able to delete the rectangle in the right pane, could "erase active shape" be enabled for this pane?

emmanuelle:
Thanks @nicholas-esterer ! I updated the modebar buttons so that the two views are more symmetric (one is for closed path and the other one for a rectangle, but shapes can be erased in both figures).

emmanuelle:
anything else I should do here @nicholas-esterer ?

nicholas-esterer:
One small thing might be to reset the found 3D shape (display no red) if one of the shapes changes. This will sort of demonstrate that the user needs to select the histogram values again (because they changed) and that the current selections don't represent the 3D shape displayed. Or you could just write that in one of the "?" parts. But this is a minor thing, the app is working really well! I'll :woman_dancing: this in case you want to get it merged ASAP :-)

nicholas-esterer (approved):

nicholas-esterer (approved):
That's looking great! My only nitpicky comment would be that although the red is erased on the 3D figure when the shapes are changed, the red is not erased on the 2D figure whose shape was not edited. So that would be a minor thing to improve. But otherwise it is really great! :woman-dancing:

Issue #456, opened 2020-07-07
In trainable segmentation, it is very slow to add annotations after a first segmentation has been performed


emmanuelle:
# App name

https://dash-gallery.plotly.host/dash-image-segmentation/

# Description of bug

After an initial segmentation has been performed, one often wishes to add new annotations to correct the model, but the loading component seems to be spinning forever and it is not possible to add new annotations. Unchecking "Show segmentation" does not seem to solve the problem.

Tagging you here @nicholas-esterer, I can also take a look after we decide together on priorities for the CZI apps. 

Ideally it should not take more time to add new annotations than before the segmentation has been computed.

![segmentation-bug](https://user-images.githubusercontent.com/263366/86804424-33262300-c077-11ea-83ea-f5da6e56f38e.gif)

nicholas-esterer:
It's true adding additional shapes makes annotating take a little bit longer. But running locally every time I add a shape it keeps working, and I can turn on/off show segmentations. I guess what I'm trying to say is it seems to be something to do with when it is deployed and so I'd like to discuss how we might go about diagnosing this?

emmanuelle:
Ah interesting, thanks for raising this point. I confirm that it runs smoothly locally. About diagnosis, is your question:
- how can we diagnose where the problem comes from now that it exists? If it's fast locally and slow on the network, my guess is that the bottleneck is the transfer of data? The only way I know is to read the callback code and to see what changes in terms of Input and Output after a first segmentation.
- or, how can we develop so that we work always with a deployed version so that we can catch such problems early enough and not at the end? For this, the solution can be do deploy very frequently to playground (probably by pushing commits to a branch here on dash-sample-apps), or to work with Dash Enterprise workspaces so that the development is directly on the server and you always check a deployed version.

nicholas-esterer:
I guess it is both, but I only have insight for the first kind of problem, haha (I never thought about DE workspaces though, that might help, although I'd prefer to keep using the tools I use normally if possible :) ).
I can think of a couple of places where we can save bandwidth
 - right now the image is uploaded to the client every time the annotations change. For the 3D app I figured out how to make that unnecessary by taking the "background" image from a store and just uploading the segments. So we could use a similar approach in this app. The background image would be loaded once upon initial page load.
 - the training doesn't use any prior information when you add new points (it just starts the training from the beginning). I remember you saying that it could be possible to use the old training and just add the new shapes, and continue training?

emmanuelle:
[...]

PR #458, opened 2020-07-07
Modal trainable segmentation

emmanuelle:
I have added a pop-up with more explanations (inspired by the CZI blog post).

Edit: while I was at it, I tried to improve the performance when adding new annotations after a segmentation, see #456 

# App pull request
- [ ] This is a new app

## About

- Current gallery app URL:  https://dash-gallery.plotly.host/dash-image-segmentation/

## Workflow

  `assets/` folder.

## The pre-review review


apps/dash-image-segmentation/app.py
```diff
@@ -322,9 +370,10 @@ def annotation_react(
     sigma_range_slider_value,
     masks_data,
     segmentation_data,
-    classifier_store_data,
```
emmanuelle, line 325: these two objects are either recomputed from scratch or not changed so they are not needed as State. Removing them reduces the amount of data to be shipped over the network, resulting in a significant performance improvement as can be seen on the playground deployment.

nicholas-esterer:
Sorry I didn't know you wanted my review (I wasn't added as reviewer). Yes, I tried the app and it still works for me. It's true the classifier_store_data can be removed. It was being passed in before because I wanted to cache old segmentations so they could just get looked up when a shape was deleted, but I abandoned this because the store would keep growing in size. So, yes :woman_dancing:

emmanuelle:
Oh interesting. It's true that we have a trade-off between the computation time on the server and the network transfer time, and if we cache large objects in a Store it can be faster to recompute stuff than pass it back and forth between client and server. Merging then!

PR #461, opened 2020-07-17
Optimize trainable segmentation

emmanuelle:
# App pull request
- [ ] This is a new app

## About

- Playground deployment URL (new version): https://dash-gallery.plotly.host/dash-image-segmentation/
- Current gallery app URL: 

This PR proposes several improvements to the trainable segmentation app. Changes are organized in 4 commits which are (I think) independent so it's possible to accept only some of them.

- make the code compatible with single-channel images (the image here is RGB but if someone wants to download the code and put another image it will work with grayscale images as well)
- computing features as soon as possible so that they are already computed when the user checks the "Show segmentation" button.
- do not update the stores for the classifier and segmented image every time as new segmentation is computed, in order to accelerate the reactivity of the app.The classifier in particular can be as large as a few Mo. The solution I found for this is to introduce a new Button "Save results" which needs to be pressed before pressing the Download buttons. It's not very elegant (in addition the segmentation is recomputed) but it saves quite a lot of time during the interactive work and I think we should give precedence to interactive work (annotating and then refining the segmentation). There is probably room for improvement for this specific change.
- setting the uirevision so that it's possible to keep a given zoom level or pan position when annotating


apps/dash-image-segmentation/trainable_segmentation.py
```diff
@@ -0,0 +1,289 @@
+from itertools import combinations_with_replacement
```
emmanuelle, line 1: this is the module I submitted to a scikit-image PR, it's more optimized than the previous version and it has a more flexible API.

apps/dash-image-segmentation/app.py
```diff
@@ -329,21 +336,37 @@ def show_segmentation(image_path, mask_shapes, segmenter_args):
...
         mask_shapes,
         img_path=image_path,
-        segmenter_args=segmenter_args,
         shape_layers=shape_layers,
         label_to_colors_args=label_to_colors_args,
+        features=features,
     )
     # get the classifier that we can later store in the Store
-    classifier = save_img_classifier(clf, segmenter_args, label_to_colors_args)
+    classifier = save_img_classifier(clf, label_to_colors_args)
```
emmanuelle, line 344: I need to find a way to put back the `segmenter_args`, since it's needed to compute features but I thought I'd submit first an early version of this PR to your review.

nicholas-esterer:
Ok I see what you want to do with the `features_dict`. I think it's a good idea but I think we'll have problems if it is done this way because we can't guarantee which process's dict we're referring too. How about we use a persistent database that stores hashes based on some client id: the processing gets done in the background on the server and the segmentation and classifier is stored in the persistent DB. Then when we want to save / store it's recalled from this DB.

emmanuelle:
closing in favor of #462

PR #468, opened 2020-07-22
joblib trainable seg: fixed script regression and integrated "Save Classifier" into existing buttons.

nicholas-esterer:
Makes it possible to run the classifier the app "learned" on an image locally with the included script `use_ml_image_segmentation_classifier.py`.
Makes it so you don't need the "Save Classifier" button: pressing the "Download classifier" or "Download classified image" button will then fill the respective `dcc.Store` with the classifier or classified image and then make a blob and make this available for download. Otherwise `dash.no_update` is passed to the `dcc.Store`s (like before).


emmanuelle:
Thanks for cleaning up my workarounds :-). I'll merge this PR into my branch and check on playground that everything works fine.

PR #477, opened 2020-08-12
Added dash-3d-image-partitioning

nicholas-esterer:
# App pull request
- This is a new app

3d image partitioning app, see https://github.com/plotly/dash-3d-image-partitioning

## About

- Playground deployment URL (new version): https://dash-playground.plotly.host/dash-3d-image-partitioning/
- Current gallery app URL: (delete this line if inapplicable): https://dash-gallery.plotly.host/dash-3d-image-partitioning/

## Workflow

- [ x] I have created a branch in the appropriate monorepo, and the

## The pre-review review

- [x ] Does everything in my code serve some purpose? (I have removed
- [x ] Does everything in my code have a clear purpose? (My code is


apps/dash-3d-image-partitioning/setup
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+git submodule init && git submodule update
```
emmanuelle, line 3: I think you don't need this line since you included the files which used to be in the submodule in the PR? (thanks for doing this BTW, it will easier for users I think)

apps/dash-3d-image-partitioning/plotly-common/README.md
```diff
@@ -0,0 +1,16 @@
+Stuff I use with `plotly.py` and `dash` webapps.
```
emmanuelle, line 1: I think these instructions are not necessary since the files are included in the PR?

apps/dash-3d-image-partitioning/plotly-common/dash_utils.py
```diff
@@ -0,0 +1,22 @@
+import dash
+import os
+
+
+def new_dash_app(f, assets_in_cwd=True, verbose=False, **dash_args):
```
emmanuelle, line 5: this function is not used (unless I'm missing something), can you please remove the file?

apps/dash-3d-image-partitioning/plotly-common/debug_utils.py
```diff
@@ -0,0 +1,24 @@
+import plot_common
+import image_utils
+import numpy as np
+
+
+def test_image_figure(shape=(300, 500), color="#002EA7"):
```
emmanuelle, line 6: same here this function is not used

apps/dash-3d-image-partitioning/app.py
```diff
@@ -0,0 +1,935 @@
...
+# scaled differently?
+hwscales = [(2, 2), (2, 2)]
+# the number of dimensions displayed
+NUM_DIMS_DISPLAYED = 2  # top and side
+# the color of the triangles displaying the slice number
+INDICATOR_COLOR = "DarkOrange"
+DISPLAY_BG_COLOR = "darkgrey"
+
+# A string, if length non-zero, saves superpixels to this file and then exits
+SAVE_SUPERPIXEL = get_env("SAVE_SUPERPIXEL", default="")
```
emmanuelle, line 39: here and the next line, calling directly `os.environ.get("SAVE_SUPERPIXEL", default="")` would work and would be easier to understand when reading the code.

apps/dash-3d-image-partitioning/app.py
```diff
@@ -0,0 +1,935 @@
...
+        segb, colormap=["#000000", "#E48F72"], alpha=[0, 128], color_class_offset=0
+    )
+    return (segl, seg)
+
+
+def make_default_figure(
+    images=[],
+    stroke_color=DEFAULT_STROKE_COLOR,
+    stroke_width=DEFAULT_STROKE_WIDTH,
+    shapes=[],
```
emmanuelle, line 62: the `shapes` parameter is not used in this app, could you please remove it and the corresponding code?

apps/dash-3d-image-partitioning/app.py
```diff
@@ -0,0 +1,935 @@
...
+                                    style={"width": "12.5%"},
+                                ),
+                                html.Button(
+                                    "Redo",
+                                    id="redo-button",
+                                    n_clicks=0,
+                                    style={"width": "12.5%"},
+                                ),
+                            ],
+                            style={"display": "flex", "margin": "2px 0 2px 0"},
```
emmanuelle, line 331: is it useful to have css both in the app file and in specific files in assets?

apps/dash-3d-image-partitioning/app.py
```diff
@@ -0,0 +1,935 @@
...
+        side_figure = image_display_figures_[1],
+        top_figure = image_display_figures_[0],
+        d=3,
+        sizex, sizey;
+    // append shapes that show what slice the other figure is in
+    sizex = top_figure.layout.images[0].sizex,
+    sizey = top_figure.layout.images[0].sizey;
+    if (top_figure.layout.shapes) {{
+        top_figure.layout.shapes=top_figure.layout.shapes.concat([
+            tri_shape(d/2,sizey*image_select_side_value/found_segs_data[1].length,
```
emmanuelle, line 519: I don't understand what the `tri_shape` function does here, could you please explain?

emmanuelle:
Hey, thanks for the app! I read through most of the code and left a couple of comments. Mostly my comments aim at simplifying the code and reducing its global size, since the total amount of code lines is quite impressive ! Some of the code will be simplified when we can use the new imshow with binary strings, so it's not the right time to do a lot of work simplifying the code but if they are some low-hanging fruits (like removing code which is not used at all), let's do it.

Also I did not understand how the `tri_shape` function is used.

Some possible improvements which can either be done now or left for later:
- merge all superpixels of the background (set them to the same label, they have an average grey level which is very small so this is a way to identify them) so that there are no boundaries in the background, this would be more pleasant visually. This can be done as preprocessing and saved in the superpixels file.
- probably the serialization time of the `draw_shapes_react` callback could be reduced by putting slices in a dict/object (with slice index as key) structure rather than a list, ie you would not create a slice for all indices, just the non-blank ones.

I'm really fine leaving this two points for further PRs.

apps/dash-3d-image-partitioning/app.py
nicholas-esterer: Ah yes, it returns a shape that looks like a triangle, I'll document it better.

apps/dash-3d-image-partitioning/app.py
nicholas-esterer: Yes and no, for instance, we could get away with writing all the layout in the `app.py` file. But for things like fonts it's easier to put them in `assets/style.css`. Personally I found it a little bit easier to put the CSS related to positioning in the `app.py` file because then it was possible to compute offsets, set variables at the top of the app, etc., which I found more intuitive than writing it in a separate CSS file. But I could move everything over to a separate file if it looks cleaner (some calculation is possible in pure CSS).

apps/dash-3d-image-partitioning/app.py
nicholas-esterer: :+1:

apps/dash-3d-image-partitioning/app.py
nicholas-esterer: Cool! I didn't know about the `get` method, I've been reusing this clumsy `get_env` routine for years.

nicholas-esterer:
I agree the superpixel boundaries look like MSDOS. I experimented with using vectorization to make them smooth. Then they would be SVG shapes and they could get filtered out of shapes using their color or something (so we don't confuse them with annotations). Or did you have another idea? I stopped looking into the SVG route because I couldn't find a python library that did require installing extra stuff with `apt-get`, but I guess there's this experimental `Aptfile` for heroku-like apps (which I didn't try).

I also agree we should just send the changed slices. This wouldn't be so hard to do.

emmanuelle:
Hey @nicholas-esterer I propose that you merge this PR and add it to the gallery since the functionality is awesome and I've already been showing the app to a lot of people :-). I'll tweak a few things later probably (no superpixels outside of the brain) but let's get this app published :-).

The only thing which I really did not understand was this `tri_shape` function, could you please explain?

:woman_dancing:

nicholas-esterer:
Ah yes! I added a doc for that in some commit but it's there now, maybe still cryptic though: https://github.com/plotly/dash-sample-apps/blob/47de0cb202ff6d70636923357855e44126b25c6d/apps/dash-3d-image-partitioning/app.py#L521

emmanuelle:
@nicholas-esterer please go ahead and merge the PR; it's just that I don't see which triangular shape you are talking about, could you please share a screenshot?

nicholas-esterer:
Haha, that's because they are very small. They are the triangles that show where you are in the slice stack
triangle

emmanuelle:
Thanks for the explanation. It was not blocking the PR anyway but once again :woman_dancing:

nicholas-esterer:
Sorry for all this back and forth but I think you have to review and select the button "Approve", otherwise it blocks me from merging
blocked

emmanuelle (approved):
Thanks :-)

PR #486, opened 2020-09-02
Bounding box app: demo using imshow with binary_string / source

emmanuelle:
This PR shows how to modify existing apps on image processing in order to use imshow with the "fast mode". Note that binary_string is True by default for RGB images so it is not specified here.


xhlulu:
Really excited to try out the improved `px.imshow`! 

Also sorry about the conflicts; i accidentally modified the `config.yml` on master rather than on my branch. oops!

emmanuelle:
Closing favor of #498

PR #493, opened 2020-09-04
Fixed uninitialized variable bug in update_ternary_map

lpyeh:
I experienced an error when running the dash-oil-gas-ternary dash app. There was a variable (prop_id) in update_ternary_map() that was being used without code to account for the times it wasn't initialized. I just return the current figure if the map doesn't need to be updated. It's a small fix, but I didn't get the error again (or other bugs) when the page updated.


xhlulu (approved):
Good catch! :woman_dancing: once the tests pass

PR #496, opened 2020-09-11
Fix layout declaration for dashr-ncbi-explorer

rpkyle:
This PR proposes to update `app$layout` such that the specification of components is contained with a list, which is wrapped within `htmlDiv`. The previous layout definition was supported prior to plotly/dashR#121, but now results in an error. The changes should restore the app's functionality.


rpkyle:
Successfully deployed to dash-playground with required fixes.

HammadTheOne (approved):
Looks good to me! :woman_dancing:

Issue #499, opened 2020-09-23
Add modal with some explanations to bounding box app


emmanuelle:
The [modal component of `dbc`](https://dash-bootstrap-components.opensource.faculty.ai/docs/components/modal/) could be used here.  

Also, one could update the code of the trainable segmentation app to use dbc's modal, since the code is not very nice at the moment.

emmanuelle:
@surchs this could be a good issue to get started.

surchs:
Closed in #506

Issue #457, opened 2020-07-07
Add link to source code to trainable segmentation and bounding box app


emmanuelle:
See for example https://dash-gallery.plotly.host/dash-volcano-plot/ and the "view on github" button.

emmanuelle:
@surchs another good issue to get started.

surchs:
@emmanuelle as you pointed out, the "view on github" functionality comes from [layout_helper](https://github.com/plotly/dash-bio/blob/81bb6fa257febb59d7841f8c5573e7231f5a9095/tests/dashbio_demos/common/layout_helper.py#L9) in `dash-bio-demo` and makes a number of assumptions on where to find the source code based on the app name. I don't think there is something similar for `dash-sample-apps`. 

I could make this link explicitly in the app but we would then have to copy it to other apps if they should also have this feature. Do you think we should make a shared module for the layout of the `dash-sample-apps` as well?

surchs:
Closed in #506

PR #506, opened 2020-10-06
Bounding box information

surchs:
Issue for app: #457 and #499

# Summary
- added a link to the github repo
- added a modal overlay with the readme information
- added dash-bootstrap to the requirements

# App pull request
- [ ] This is a new app
- [X] I am improving an existing app (redesigns/code "makeovers")

## About

- Playground deployment URL: https://dash-playground.plotly.host/dash-image-annotation/
- Current gallery app URL: https://dash-gallery.plotly.host/dash-image-annotation/
- Python app repository link: https://github.com/plotly/dash-sample-apps/blob/master/apps/dash-image-annotation/README.md

## Workflow

- [ ] If the app is a redesigned and/or restyled version ofI wen an
  `assets/` folder.

## The pre-review review

- [ ] Am I reinventing the wheel? (I think not. There is the [layout_helper](https://github.com/plotly/dash-bio/blob/81bb6fa257febb59d7841f8c5573e7231f5a9095/tests/dashbio_demos/common/layout_helper.py#L9) module in the dash-bio apps which adds a pretty github link. But here we went with a button)


emmanuelle:
Thank you for the PR! I really like the fact that you could add the two buttons in the header without too much css, which was the idea of using dbc and it seems to work out well (once you had solved the conflicts with another stylesheet).

About the modal, using the Readme is a good idea, I hadn't thought about this. Would it be possible to use a markdown file with info more specific about what users might expect when they click on the button? I think they should be told that they can draw rectangles over different types of objects, select the object class in the dropdown or in the table, and switch between different images.

Also there seems to be a small rendering glitch when opening the markdown file, every line break in the file becomes a `<br>` in the html which causes unwanted line breaks. There might be an option in `dcc.Markdown` to remove line breaks or it can be done with string parsing.

apps/dash-image-annotation/app.py
```diff
@@ -182,8 +190,60 @@ def shape_data_remove_timestamp(shape):
...
+                        ),
+                        dbc.Col(
+                            html.H1("Bounding Box Classification App", id="title"),
+                            width=4,
+                            align="center",
+                        ),
+                        dbc.Col(
+                            [
+                                dbc.Button(
+                                    "Readme",
```
emmanuelle, line 210: What would you think of using "about" or "learn more" instead of "readme"? I think "about" has my preference but it's not a strong one.

surchs: Yes, both are better. I'm leaning a bit more to "learn more" as it seems more explicit to me.

surchs:
I like the idea of a dedicated markdown file for the modal. 

I think the additional linebreaks came from me using `.readlines` rather than `.read`. Somehow `dcc.Markdown` adds newline characters between the list elements created by `.readlines`. So reading it into a single string removes this behaviour.

emmanuelle:
Great, thanks for investigating this markdown issues. Indeed `.read` should work better. 

About the github "this branch is out-of-date with the base branch", you can merge master into your branch by clicking on "update branch" (then don't forget to pull from origin on your local branch before you push new commits).

apps/dash-image-annotation/assets/Howto.md
```diff
@@ -0,0 +1,18 @@
...
+- export the annotations you have made
+
+## How to use this app
+To annotate the image, first select the annotation label you want to apply: 
+![Screenshot of label selector](assets/select_label.png)
+
+Then draw a rectangle on the image area you want to annotate with your cursor:
+![Screenshot of bounding box](assets/draw_annotation.png)
+
+You can continue annotating additional areas of the image, selecting a different image with the `previous` and `next
```
emmanuelle, line 16:
```suggestion
You can continue annotating additional areas of the image, selecting a different image with the `previous` and `next`
```

emmanuelle (approved):
Thanks for the modifications! I just left a suggestion since I believe there's a formatting typo, but otherwise :woman_dancing: !

surchs:
Yay :woman_dancing: ! I believe I can only merge this once you have approved the review @emmanuelle

emmanuelle (approved):
Ha ha yes I always forget

Issue #501, opened 2020-10-02
Try to use dbc components in CZI apps


emmanuelle:
# App name : bounding box app, trainable segmentation app, superpixels app

At the moment we use quite a lot of CSS to make sure that the different parts of the app are reactive (when the screen is resized etc.). It could be interesting to try to use dash-bootstrap-components to see if the resulting code is easier to read.

emmanuelle:
See for example https://dash-bootstrap-components.opensource.faculty.ai/docs/components/card/

xhlulu:
Awesome! I've used `dbc` in a few recent apps, their components makes it easy to use [various themes](https://dash-bootstrap-components.opensource.faculty.ai/docs/themes/) without changing CSS.

surchs:
I think the code gets more readable with dbc and using these components seems generally easier than going the css route. An added benefit is that we can probably remove a lot of the css in the custom stylesheets and thereby improve their readability as well. 

I have also taken some pointers from @xhlulu and broke the layout up into separate components like in the [aix360-heart](https://github.com/plotly/dash-sample-apps/blob/master/apps/dash-aix360-heart/app.py). I think this also improves readability a lot.

PR #511, opened 2020-10-16
Layout image-annotation-app with dash-bootstrap-components and reduce css

surchs:
Issue for app: #501 

# App pull request
- [ ] This is a new app

## About

- Playground deployment URL: https://dash-playground.plotly.host/dash-image-annotation/
- Current gallery app URL: https://dash-gallery.plotly.host/dash-image-annotation/

## Workflow

  `assets/` folder.

## The pre-review review


apps/dash-image-annotation/app.py
```diff
@@ -185,165 +181,236 @@ def shape_data_remove_timestamp(shape):
...
                     figure=fig,
                     config={"modeBarButtonsToAdd": ["drawrect", "eraseshape"]},
+                )
+            ]
+        ),
+        dbc.CardFooter(
+            [
+                dcc.Markdown(
+                    "To annotate the above image, select an appropriate label on the right and then draw a "
+                    "rectangle with your cursor around the area of the image you wish to annotate\n\n"
```
nicholas-esterer, line 233:
```suggestion
                    "rectangle with your cursor around the area of the image you wish to annotate.\n\n"
```

nicholas-esterer (commented):
Just a small punctuation suggestion. Also for some reason I thought it was possible at some point to change the annotation type in the table (say change "tree" to "road") but it doesn't seem the old one does this either. In any case the possible entries are initialized when initializing the table (see lines 277-285). So I'm wondering if this went away at some point or just never worked and was forgotten about, haha.
Also the timestamp table is missing, but maybe you've decided to let go of that, which is cool. Just wanted to make sure it wasn't forgotten :).
Otherwise, looks really good :+1:, must more consistent than what we had before.

surchs:
Thanks for the suggestion! Yeah, I also noticed that the table dropdown didn't seem to allow changing the label. I think that didn't work in the first version I started working on. But would probably be a good idea to get to work. Maybe for another PR?

I did get rid of the timestamp table. Since it was a separate data-table object, I had a very hard time keeping it in the same line as the annotation table  when resizing the view port. The data are still available in the .json download so I opted to remove the timestamp table to avoid strange layout breaks.

emmanuelle (approved):
wow impressive work! Thanks, it works out really well with `dbc`. I don't see anything to change so :dancer: !

PR #516, opened 2020-10-27
Color palette is now more colorblind-friendly

### LINKED ISSUE #450: Use colorblind-friendly sequence of colors for trainable segmentation app
# App name

https://dash-gallery.plotly.host/dash-image-segmentation/

# Description of bug

The sequence of colors used for annotations is not colorblind-friendly. The `Plotly` or `d3` sets of colors in https://plotly.com/python/discrete-color/ would be better

nicholas-esterer:
I chose a subset of the Wong palette described here:
https://www.nature.com/articles/nmeth.1618
closes #450


apps/dash-image-segmentation/app.py
```diff
@@ -32,7 +32,7 @@
 # the number of different classes for labels
 NUM_LABEL_CLASSES = 5
 DEFAULT_LABEL_CLASS = 0
-class_label_colormap = px.colors.qualitative.Light24
+class_label_colormap = ["#E69F00", "#56B4E9", "#009E73", "#F0E442", "#0072B2"]
```
surchs, line 35: Since we only have 5 categories, maybe you could use the "black" color tone from the paper. It seems to be the most distinct from the others.

surchs (approved):

surchs:
Looks good, one small comment on the color choice. Otherwise :woman_dancing:

emmanuelle:
Thanks @nicholas-esterer ! Looks good to me, :woman_dancing: 

@surchs the black color is indeed very distinct but I'm not sure it renders that well for semi-transparent overlays. I'll let Nick decide whether he wants to try it out ;-).

emmanuelle (approved):
Thanks :-)

nicholas-esterer:
It's true that black is more contrasting and it works when changing its alpha channel for a segmentation image
image
but to me it seemed sort of out of place maybe? For example if there were text on the image it would probably be in black. Also I hoped to solve the problem that the yellows and blues are sort of similar, but it seems it isn't possible for 5 colors, even if you include the black (this website tries to simulate the effect).
https://davidmathlogic.com/colorblind/#%23000000-%23E69F00-%2356B4E9-%23009E73-%23F0E442-%230072B2-%23D55E00-%23CC79A7
So probably I will just leave it as is :-)

Issue #507, opened 2020-10-07
Port label properties app to this repo


emmanuelle:
A new app for the gallery should be adapted from https://github.com/plotly/canvas-portal/blob/master/apps/object-properties/app.py, with the following improvements:

- the existing code does not work with recent versions of `dash_table`, it should be ported to the new API
- the app was written at a time where `px.imshow` did not exist, using `imshow` should make the code simpler and more pleasant to read
- the app should be styled to have a visual appearance similar to existing apps of the gallery

refactoring this app is also a good time to discuss its specifications. At the moment the interactivity consists in highlighting table rows when hovering over the image, and highlighting parts of the image when selecting some table rows. However, it would also be possible to add some information about labels in the hover information of the figure.

emmanuelle:
Closed by #518.

PR #518, opened 2020-10-28
Add label properties app

surchs:
Issue for app: #507 

# App pull request
- [X] This is a new app

## About

- Playground deployment URL (new version): https://dash-playground.plotly.host/dash-label-properties
- Old Python app repository link: https://github.com/plotly/canvas-portal/blob/master/apps/object-properties/app.py

## Workflow

- [X] I have created a branch in the appropriate monorepo, and the
- [X] If the app is a redesigned and/or restyled version of an
- [X] If the app is on the Dash Gallery portal, I have added a link to
- [X] I have removed *all Google Analytics code* from the app's
  `assets/` folder.

## The pre-review review


surchs:
This PR isn't completely finished yet. For example, the app currently doesn't show very interesting hover information as discussed in #507. I wanted to start the review process at this stage already because there are a couple of changes that I'd like feedback on.

apps/dash-label-properties/app.py
```diff
@@ -0,0 +1,357 @@
...
+
+@app.callback(
+    Output("table-line", "style_data_conditional"), [Input("graph", "hoverData")]
+)
+def higlight_row(string):
+    """
+    When hovering hover label, highlight corresponding row in table,
+    using label column.
+    """
+    if not dash.callback_context.triggered:
```
surchs, line 272: I have added this callback check (and a similar one in line 319) to "catch" None-type errors right when the app starts. In the original app, there was a `suppress_callback_exceptions` that I have removed and that I believe was intended to take care of this. Maybe there is a better way of dealing with these issues

apps/dash-label-properties/app.py
```diff
@@ -0,0 +1,357 @@
...
+        img = PIL.Image.fromarray(img)
+    labels = labels.astype(np.float)
+    custom_viridis = colors.PLOTLY_SCALES["Viridis"]
+    custom_viridis.insert(0, [0, "#FFFFFF"])
+    custom_viridis[1][0] = 1.0e-4
+    # Contour plot of segmentation
+    print("mode is", mode)
+    opacity = 0.4 if mode is None else 1
+
+    fig = px.imshow(img, binary_backend="jpg")
```
emmanuelle, line 94:
```suggestion
    fig = px.imshow(img, binary_string=True, binary_backend="jpg")
```

emmanuelle: since the image is grayscale you need to pass `binary_string=True`, the default is `False` for single-channel images.

apps/dash-label-properties/app.py
```diff
@@ -0,0 +1,357 @@
...
+    fig.add_contour(
+        z=labels,
+        contours=dict(start=0, end=labels.max() + 1, size=1, coloring=mode),
+        line=dict(width=1),
+        showscale=False,
+        colorscale=custom_viridis,
+        opacity=opacity,
```
emmanuelle, line 101: a hovertemplate could be added here to display only the `z` value (not `x` and `y`) and to add for example the mean intensity (passed as `customdata`, see for example https://plotly.com/python/hover-text-and-formatting/#adding-other-data-to-the-hover-with-customdata-and-a-hovertemplate)

apps/dash-label-properties/app.py
```diff
@@ -0,0 +1,357 @@
...
+    )
+    # Remove axis ticks and labels
+    fig.update_xaxes(showticklabels=False).update_yaxes(showticklabels=False)
+
```
emmanuelle, line 105: could you please also decrease the margins here? The image is quite small compared to the total figure area.

apps/dash-label-properties/app.py
```diff
@@ -0,0 +1,357 @@
...
+app = dash.Dash(__name__, external_stylesheets=external_stylesheets)
+server = app.server
+
+# Read the image that will be segmented
+filename = (
+    "https://upload.wikimedia.org/wikipedia/commons/a/ac/Monocyte_no_vacuoles.JPG"
+)
+img = io.imread(filename, as_gray=True)[:660:2, :800:2]
+labels = measure.label(img < filters.threshold_otsu(img))
+props = measure.regionprops(labels, img)
```
emmanuelle, line 33: here we can now use the `regionprops_table` function of scikit-image (recent addition), see for example https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_regionprops_table.html#sphx-glr-auto-examples-segmentation-plot-regionprops-table-py The syntax will be more idiomatic for pandas users and so will (I think) be the creation of the `DataTable`

apps/dash-label-properties/app.py
emmanuelle: I'm not sure what the original intention is (remember that this was maybe the first Dash app I ever wrote so expect the code to be really bad ;-)). But we now have the possibility to pass `prevent_initial_call=True` in the arguments of `app.callback` which seems like the way to go here.

apps/dash-label-properties/app.py
```diff
@@ -0,0 +1,357 @@
...
+            "backgroundColor": "#3D9970",
+            "color": "white",
+        }
+    ]
+
+
+@app.callback(
+    [Output("graph", "figure"), Output("cache", "data"), Output("row", "children")],
+    [
+        Input("table-line", "derived_virtual_indices"),
```
emmanuelle, line 288: It looks like it's not possible to select rows (see 
image 
maybe there is something fishy with the CSS and the selection columns gets cut?
) so this input is never used

apps/dash-label-properties/app.py
```diff
@@ -0,0 +1,357 @@
...
+        dbc.CardBody(
+            dbc.Row(
+                dbc.Col(
+                    [
+                        dash_table.DataTable(
+                            id="table-line",
+                            columns=columns,
+                            data=table.to_dict("records"),
+                            filter_action="native",
+                            row_deletable=True,
```
emmanuelle, line 239: add `row_selectable='single'` or `'multi'` to be able to select columns

apps/dash-label-properties/app.py
emmanuelle: see my comment above in the definition of the table

emmanuelle: Also in this callback, I'm not sure anything happens when a table row is deleted

emmanuelle (commented):
Thanks for the PR! I left a couple of comments. Also, some columns of the table are a bit too wide, it would be possible to change columns headers (`euler_number` --> "Euler number") to fit on two lines to have less wide columns. This would make it possible to have a wider column for the figure. What would you think of this?

apps/dash-label-properties/app.py
surchs: oh nice, I'll use this instead, thanks! Haha, I wasn't sure if it was something that worked in an older version or where this was coming from. I think the errors only showed up in the console or during debug mode so they didn't interfere with the functionality.

apps/dash-label-properties/app.py
surchs: table_interactivity

- I just tried with the [playground deployment](https://dash-playground.plotly.host/dash-label-properties) and something happens when I select a row, i.e the corresponding element gets a red outline. Selecting a row is however super slow to draw the outline, I had noticed that locally as well. I'm not sure if it has to do with the fact that for every change, we re-build the figure from scratch. Is that what you expected?

- If I delete a table row, the corresponding element is no longer overlayed/highlighted. Is there something else that should happen?

- if I highlight multipe rows (e.g. by shift-click), only the first row gets highlighted. It would be nice to have all regions annotated I think.

emmanuelle: You are right, the behaviour is correct (selecting one row or deleting rows has the expected behaviour), I just was not patient enough to see the result :-). Performance is indeed an issue here. We can think of several ways of reducing the size: 
- passing the colored overlay as a png string (which should be quite small) in an image trace. You'll have to create a multichannel array for the labels. For this you can use https://datashader.org/api.html#datashader.transfer_functions.shade. However, the down side of passing a colored image instead of a contour plot is that we will not have the index any more in the hover. For this it can be interesting to reproduce what's done in https://scikit-image.org/docs/dev/auto_examples/segmentation/plot_regionprops.html#sphx-glr-auto-examples-segmentation-plot-regionprops-py with filled scatter traces.
- passing the whole `labels` array as `State` and `Output` in the callback is suboptimal, a list of unique labels should be enough.
- finally, it's possible to keep the background image in a store, to have the computed overlay in another store which is the only one used as output in the callback, and to assemble the figure in a clientside callback. I'm hoping that the previous points will optimize enough so that we don't have to resort to clientside callbacks, let's see.

nicholas-esterer:
Sometimes I get the following error message:
```
  File "/home/nicke/development/dash-sample-apps/apps/dash-label-properties/app.py", line 299, in higlight_row
    "if": {"filter_query": "{label} eq %d" % index},
TypeError: %d format: a number is required, not dict
```
(there's more in the traceback but it's not in your app anymore)
For some reason `index` is sometimes a dict? I can't find what triggered this though, sorry!

surchs:
> Sometimes I get the following error message:
> 
> ```
>   File "/home/nicke/development/dash-sample-apps/apps/dash-label-properties/app.py", line 299, in higlight_row
>     "if": {"filter_query": "{label} eq %d" % index},
> TypeError: %d format: a number is required, not dict
> ```
Hhm, I also remember seeing that but I thought I had debugged this and haven't seen it since. Does that usually happen right at the start of the app? I think that might have been taken care off with the `prevent_initial_call=True` but I am not sure. I'll see if I can make it come up again.
> 
> (there's more in the traceback but it's not in your app anymore)
> For some reason `index` is sometimes a dict? I can't find what triggered this though, sorry!

I don't think I understand what this means. Have I removed something that makes the traceback more informative?

nicholas-esterer:
No no, it's just I can't always trigger it, and I didn't look at the problem that long (I had hoped I could show you better where the bug was). Maybe just try putting
`print(index)`
just before and see what it shows.

surchs:
Ah ok, cool, thanks for that. I have definitely also seen this but without remembering the specific fix, I thought I had taken care of it. Would be curious to see what brings it up.

apps/dash-label-properties/app.py
```diff
@@ -0,0 +1,376 @@
...
+    overlay_data = np.dstack(
+        [
+            values[:, col_id][labels.astype(int)]
+            for col_id, col in enumerate(overlay_columns)
+        ]
+    )
+    # Display hover data with precision if data is float
+    hover_string = [
+        f"{col}: %{{customdata[{col_id}]:.3f}}"
+        if data_table[col].dtype in (np.float,)
```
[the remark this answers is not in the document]
nicholas-esterer: This is basically what `np.array_str` does (but it adds these `[...]` things around the numbers), so it sort of seems the way to do it. The arrays I think are just numpy arrays underneath. I think it's ok that `data_table[col].dtype in "float"` float has the same result, because in numpy you can make an array like `np.zeros(N,dtype='float')` or `np.zeros(N,dtype=np.float)` and they will have the same type. That it doesn't check that it is one of those two types is a another story involving Python's type checking system in general... but I think we don't have to worry about that (we provided the data, haha).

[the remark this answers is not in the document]
surchs: Thanks for the pointers @nicholas-esterer! I dug a little further because I didn't understand why `np.dtype("float") == "float"` evaluates `True` essentially. 
The way I understand it now is that 1) [numpy datatypes have a hierarchy](https://numpy.org/doc/stable/reference/arrays.scalars.html) where `np.float64` is a sub-datatype of `float` and 2) numpy array's have [a `__eq__` method](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.__eq__.html) that evaluates this hierarchy internally. So you can also do things like `np.dtype("float") == "double"` and so on...
Guess I learned something new today!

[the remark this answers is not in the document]
surchs: I think to be super explicit here, I'll just go with [`np.issubdtype`](https://numpy.org/doc/stable/reference/generated/numpy.issubdtype.html) here to avoid any confusion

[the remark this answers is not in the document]
nicholas-esterer: Maybe you can get away with just using printf statements, does this do what you want?
```
>>> '%0.3f' % (1.23,)
'1.230'
>>> '%0.3f' % (1,)
'1.000'
```
1 is an int but has 0s after it, 1.23 is made to have an extra 0

surchs:
I have encountered a bit of a tricky issue: I have added `datashader` to the requirements to convert my 2D numpy array to a 3-channel RGB image. `datashader` depends on `numba` which depends on `llvmlite` which in turn depends on [`llvm`](https://pypi.org/project/llvmlite/). Now `datashader` is pretty up to date in terms of its requirements and the current version requires `llvm`>8 to install. Our CircleCI container is running Debian Stretch which offers up to `llvm-7` (including backports). There is a [nightly repo for llvm](https://apt.llvm.org/) that provide recent versions and I can successfully install and configure it in the CI. However, even with this set up, installing `llvmlite` eventually fails because it can't find `SVML`. 

I don't really understand why this is proving so difficult. After all there is a `data-shader` app in this repo and the [config doesn't appear to be radically different](https://github.com/plotly/dash-sample-apps/tree/c36b89a4c87a5f255f48c3c46c799e883344486f/apps/dash-datashader) (other than pinning python 3.6). But bottom line: I can't manage to get the app with datashader deployed through the CI. Maybe @nicholas-esterer or @emmanuelle, you have any ideas?

emmanuelle:
@jonmmease maybe you can help @surchs with the deployment problem coming from datashader's dependencies.

@surchs if this is blocking you can probably vendor datashader's module together with the app, keeping only what you need and removing the numba decorator of https://github.com/holoviz/datashader/blob/master/datashader/transfer_functions/__init__.py

Before this you can try to also pin numba to an older version explicitly, probably pinning only datashader is not enough.

jonmmease:
I touched base with @surchs offline a bit on Friday. Here are a set of versions that currently work with the old version of pip that is included in the DE buildpack (https://github.com/plotly/dash-world-cell-towers/blob/d02ebf370b6f31196386a1939dc5e3ac3fe71317/requirements.txt).

After pip is updated (https://github.com/plotly/heroku-buildpack-python/pull/38), these errors should all go away.

surchs:
Thanks @jonmmease for the pointers to the working versions and @emmanuelle for the ping. I will try these out with the next commit!

surchs:
I believe that the core functionality is there. Remaining issues:
- [CANT FIX] colorbar is not consistent through filter events because datashader.shade doesn't implement `span` correctly in this version (to manually set the value range)
- [HELP WANTED] the hidden (opaque) scatter traces used to get the hoverinfo somehow add margins to the canvas that look annoying and also disappear when user zooms in. Maybe there is some elegant way to solve this @nicholas-esterer ?
- [WILL FIX] we should probably update the modal with the new functionality and new screenshot.

jonmmease:
> [CANT FIX] colorbar is not consistent through filter events because datashader.shade doesn't implement span correctly in this version (to manually set the value range)

FWIW, span doesn't work with the default shade method (`shade(..., how=eq_hist)`), but it should work with the other methods (`linear`, `log`, and `cbrt`) unless there was a bug in this version of datashader.  These won't give the same even distribution of colors, but you may want to try them out to get span support.

Code: https://datashader.org/_modules/datashader/transfer_functions.html#shade

emmanuelle:
Thank you @surchs I really love the app! The dropdown with the different ways of coloring the objects is very nice. One thing I noticed is that if you want to select an object in the table (the red outline) you need to select a value in the dropdown first or it won't work. I did not investigate why, I can take a look unless it's obvious for you.

Also could you please explain
> [HELP WANTED] the hidden (opaque) scatter traces used to get the hoverinfo somehow add margins to the canvas that look annoying and also disappear when user zooms in. Maybe there is some elegant way to solve this @nicholas-esterer ?
I did not notice annoying margins. 

Last thing (before I read the code, I've just tested the deployed app for now): I'm not sure I understand what the checkboxes in the columns headers are for (maybe it's this cell selection thing I was talking about and I was not clear about what I wanted? Anyway for now the app has all the features I asked for, but I don't know how to use these checkboxes). 

And yes please write one of these great modals you've been treating us with, with a screenshot of the much-improved layout :-).

apps/dash-label-properties/requirements.txt
```diff
@@ -0,0 +1,16 @@
+datashader==0.8.0
+dash-core-components
```
emmanuelle, line 2: I thought that when you put `dash` in the `requirements.txt`, or pip install dash, you got dash-core-components, dash-html-components and dash-table together with `dash`. Am I missing something? (maybe it's related to the datashader problems?) I'm asking because we've been trying to tell users (on the forum for example) to install `dash` only and not to separately install dcc and html because they would run into version incompatibilities.

apps/dash-label-properties/assets/object_properties_style.css
```diff
@@ -0,0 +1,9 @@
+/* set the margins of row elements to zero to
+override the settings in the dash bootstrap stylesheet */
+.row {
```
emmanuelle, line 3: I love it when the CSS stylesheet has a length of 9 lines :-).

apps/dash-label-properties/app.py
emmanuelle: ah ok it's the hover info which depends on the selected columns. Got it!

emmanuelle: and very cool :-). Maybe just add a hint somewhere to explain this.

emmanuelle:
ok I saw the margins
image

emmanuelle:
For the margins a quick fix would probably to use a different template such as `simple_white` which has a white background and no grid? https://plotly.com/python/templates/ There might be a more elegant solution of course.

surchs:
> Also could you please explain
> 
> > [HELP WANTED] the hidden (opaque) scatter traces used to get the hoverinfo somehow add margins to the canvas that look annoying and also disappear when user zooms in. Maybe there is some elegant way to solve this @nicholas-esterer ?
> > I did not notice annoying margins.
> 
image
I mean these "margins" at the top and bottom of the figure. If you zoom in, they disappear: 
image

They are related to the [hidden scatter trace that the hover-info is based on](https://github.com/plotly/dash-sample-apps/pull/518/files#diff-b49b83ea7fe4ab7dc900eb4ae2b04bc624d88af9295822b231267701f083121fR143) (i.e. if the traces are not added, the figure fills the entire canvas). I assume that the scatter traces add their own x and y range "padding" and therefore extend beyond the base image range. So far I haven't found a good way to keep them from doing that. Maybe I should just manually compute the max/min of the positions and set the ranges to that?

> Last thing (before I read the code, I've just tested the deployed app for now): I'm not sure I understand what the checkboxes in the columns headers are for (maybe it's this cell selection thing I was talking about and I was not clear about what I wanted? Anyway for now the app has all the features I asked for, but I don't know how to use these checkboxes).
> 

Yeah, I need to explain this :). I was thinking what we could do with the selected_columns feature and so at the moment it controls the information that is shown in the hover-info. That is, you only see the variables from the selected columns (because by default I showed all and that gets a bit busy). I tried to use a [datatable tooltip](https://dash.plotly.com/datatable/tooltips) to explain this but I think the version we use here doesn't have that feature yet? The description could either go in the modal or on top of the table inside the card.  

> And yes please write one of these great modals you've been treating us with, with a screenshot of the much-improved layout :-).

Will do :)

apps/dash-label-properties/requirements.txt
surchs: I may have just copied this over from an existing requirements.txt. I'll make sure we don't need it and then trim the requirements!

apps/dash-label-properties/app.py
```diff
@@ -0,0 +1,456 @@
...
+]
+# Select the columns that are selected when the app starts
+initial_columns = ["label", "area"]
+
+img = img_as_ubyte(color.gray2rgb(img))
+img = PIL.Image.fromarray(img)
+
+
+def image_with_contour(
+    img, active_labels, data_table, active_columns, color_column, mode="lines"
```
emmanuelle, line 69: I think in this app `mode` is always None so probably you can simplify the code a bit here?

apps/dash-label-properties/app.py
```diff
@@ -0,0 +1,456 @@
...
+    # Background cannot be zero when using datashader (because 'span' is not correctly implemented
+    # in current old version)
+    masked_values = np.ma.masked_where(~mask, value_array)
+
+    print("mode is", mode)
+    opacity = 0.4 if mode is None else 1
+
+    fig = px.imshow(img, binary_string=True, binary_backend="jpg",)
+    # Overlay the current labels
+    _, hex_list = zip(*colors.PLOTLY_SCALES["Viridis"])
```
emmanuelle, line 103: can you wrap this part (the shading) in a helper function? It will be clearer and it's also something we could reuse in other apps so it will be easier to find it back.

apps/dash-label-properties/app.py
```diff
@@ -0,0 +1,456 @@
...
+img = img_as_ubyte(color.gray2rgb(img))
+img = PIL.Image.fromarray(img)
+
+
+def image_with_contour(
+    img, active_labels, data_table, active_columns, color_column, mode="lines"
+):
+    """
+    Figure with contour plot of labels superimposed on background image.
+
```
emmanuelle, line 73: Since this function is quite advanced, would it be possible to give a few explanations in the docstring about the different traces which are used, and the tricks to improve performance?

apps/dash-label-properties/app.py
```diff
@@ -0,0 +1,456 @@
...
+                    if np.issubdtype(type(prop_val), "float")
+                    else f"{prop_name}: {prop_val}"
+                    for prop_name, prop_val in row[active_columns].iteritems()
+                ]
+            )
+            + " <extra></extra>"  # Add this to remove the trace name. See e.g. https://plotly.com/python/reference/#scatter-hovertemplate
+        )
+        fig.add_scatter(
+            x=x,
+            y=y,
```
emmanuelle, line 161: is the layout Image above really needed here, or would it be possible just to use the filled scatter traces with suitable colors computed from with datashader? Not very important since it works as it is but I'm just trying to see whether we can reduce the complexity of the function or not.

emmanuelle:
OK I finished my review!

apps/dash-label-properties/app.py
surchs: Yes, I think that's a great idea. And I think we don't even need the datashader then anymore because we can simply use the matplotlib cmap object for "viridis" and call the rgb colors for each (normalized) value we want to draw individually. I have committed this and I think it's a lot cleaner and most of the complexity is gone.

Unfortunately I don't know how to integrate the colorbar-scatter into the "color overlay" scatter. So for the moment, we have these two, relatively lengthy scatter traces in there.

apps/dash-label-properties/requirements.txt
surchs: I looked at pipdeptree and all we really need is `dash`, `dash_canvas` and `dash_bootstrap_components` (now that we got rid of the datashader dependencies). For visibility purposes it might make sense to add `skimage` back in, what do you think @emmanuelle

apps/dash-label-properties/app.py
surchs: I think we no longer need this code/function in the label-properties app. Do you think we should still wrap it into a helper function and if so, what would be a good place to put it?

emmanuelle: please remove any code that is not needed any more. In order to keep track of what you did here maybe we could add an example to the imshow page of the plotly doc (on how to pass grayscale image as a binary image string but still use a colormap) but this is quite a hack so maybe we can just remember that this is possible and we'll find the code in the PR if needed?

emmanuelle (approved):
Awesome, thank you @surchs! :woman_dancing:

PR #524, opened 2020-11-11
update app name in README.md

surchs:
minimal PR to deploy dash-label-properties-app (actually found something useful to fix)


emmanuelle (approved):
thanks :woman_dancing:

PR #525, opened 2020-11-11
Update covid-xray example to use dash-slicer

almarklein:

# App pull request
- [ ] This is a new app
- [x] I am improving an existing app

## About

- Playground deployment URL (new version): ?
- Current gallery app URL: https://dash-gallery.plotly.host/dash-covid-xray/

## Description

This modifies the Covid xray example to use the new [dash-slicer](https://github.com/plotly/dash-slicer), removing quite a bit of boilerplate code, so the example focus becomes more at the image processing logic.

Comments:
* The dash-slicer project is not yet on Pypi; the purpose of this rewrite is to tweak the API to accommodate apps like this :)
* This currently relies on features of dash-slider that are not yet in the main branch. In other words, it only really Works On My Machine (TM).
* As long as the above is the case, this stays a Draft PR.

### DIFF
--- a/apps/dash-covid-xray/app.py
+++ b/apps/dash-covid-xray/app.py
@@ -1,130 +1,77 @@
+from time import time
+
 import numpy as np
+from nilearn import image
+from skimage import draw, filters, exposure, measure
+from scipy import ndimage
+
+import plotly.graph_objects as go
+import plotly.express as px
+from plotly.express._imshow import _array_to_b64str
+
 import dash
 from dash.dependencies import Input, Output, State
 import dash_bootstrap_components as dbc
 import dash_html_components as html
 import dash_core_components as dcc
-import plotly.graph_objects as go
-import plotly.express as px
-from plotly.express._imshow import _array_to_b64str
-from nilearn import image
-from skimage import draw, filters, exposure, measure
-from scipy import ndimage
-from time import time
+from dash_slicer import VolumeSlicer
 
 app = dash.Dash(__name__)
 
-server = app.server
-
 t1 = time()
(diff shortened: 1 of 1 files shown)


apps/dash-covid-xray/app.py
```diff
@@ -1,132 +1,79 @@
...
+
+def path_to_coords(path):
+    """From SVG path to numpy array of coordinates, each row being a (row, col) point"""
     indices_str = [
         el.replace("M", "").replace("Z", "").split(",") for el in path.split("L")
     ]
     return np.array(indices_str, dtype=float)
 
 
-def make_figure(
```
emmanuelle, line 62: I'm so glad this big chunk of code can go away :-).

emmanuelle:
Thanks for opening this PR, it shows really well how the code will be simplified with dash-slicer. I've just noticed two little things:
- annotations don't persist on different slices (whereas they do in the deployed app), which is normal because the internal callback updating the figure does not know about these annotations. Since the case of persisting annotations is probably a bit specific I think I'm ok with this change but we can think a little bit if there would be some not-too-hacky way of making some annotations persist (like the overlay mask maybe?)
- I've noticed a small bug that the overlay mask is not visible everywhere it should be, like in this screenshot (I think it should be visible on the left image, according to the indicators position)
image

emmanuelle:
@almarklein I created #538 to work on a branch on `origin`. I propose to take a shot at finishing the PR now that the slicer is released and I'd like to write my blog post about the app in the next days.

almarklein:
Superseded by #538

PR #529, opened 2020-11-18
Label props better contour

surchs:
## About

- Playground deployment URL (new version): https://dash-playground.plotly.host/dash-label-properties
- Current gallery app URL: (delete this line if inapplicable): https://dash-gallery.plotly.host/dash-label-properties/

## Summary:
- fixed figure range to image shape to remove padding from scatter trace
- pad label array to get contours for the regions in the corners
- use scatter to draw contour around selected regions in green, change overlay colormap to "plasma" to avoid clash with green


emmanuelle (approved):
Thank you for the improvements!

PR #540, opened 2020-12-09
Brain connectivity app

surchs:

# App pull request
- [X] This is a new app

## About

- Playground deployment URL (new version): https://dash-playground.plotly.host/dash-brain-connectivity
- Playground (out of sync / manual): https://dash-playground.plotly.host/app-brain-conn/

## TODO:
General
- [ ] Merge the `helper_functions` into the app or a better location
- [ ] 

Functionality
- [x] Show the current slicer position in real world coordinates and voxel coordinates - like so:
image
- [ ] Allow the user to select from several (precomputed) brain Atlases
- [x] pre-compute the focus locations for each brain region so the location lies within the region (center of mass is not guaranteed to lie within)

Layout
- [ ] 

Bugs
- [x] Circular dependencies between the slicer and the other components selecting brain regions
- [ ] Have the brains in [RAS array orientation](https://nipy.org/nibabel/coordinate_systems.html#nibabel-always-uses-an-ras-output-space) show up as expected (all L-R) - like so:


apps/dash-brain-connectivity/requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+numpy
+nilearn
+nibabel
+dash-bootstrap-components
+dash-bio
+matplotlib
+pytest
+gunicorn
+scikit-image
+-e git+https://github.com/plotly/dash-slicer.git@9363a79222ccd2806c18f3753dc29327685e4bec#egg=dash_slicer
```
emmanuelle, line 14: now you can try dash-slicer 0.2.0 :-)

apps/dash-brain-connectivity/app.py
```diff
@@ -0,0 +1,510 @@
...
+                "if": {"filter_query": "{Region Name} = '%s'" % region_name},
+                "backgroundColor": "#3D9970",
+                "color": "white",
+            }
+        ]
+    return None
+
+
+if __name__ == "__main__":
+    app.run_server(debug=True, dev_tools_props_check=False, port=8052)
```
emmanuelle, line 510:
```suggestion
    app.run_server(debug=True, dev_tools_props_check=False)
```

emmanuelle: I also use this trick of changing the port because I have several apps open at the same time, but let's keep it simple for the gallery

surchs:
I have just checked the deployment, and there are still a couple of issues:

1. There seems to be no link between the dash-slicer components and the rest
2. The app loads pretty slowly

1) is most likely related to dcc not supporting the slider.drag_value attribute yet. I'll try to bundle the dev version with the deployment
I am less clear on how to go about 2). Maybe fixing 1) fixes 2) as well as there might be some waiting for errors? I'll see if fixing 1 helps and reproduce locally

surchs:
OK, looks like it deploys now.

emmanuelle:
@surchs I finally got to do a bit of debugging on the app. A first thing is that most callbacks are executed when the app is launched and this takes quite a lot of time. I would give it a try to do `prevent_initial_call=True` (https://dash.plotly.com/advanced-callbacks) and see if the app launches faster. (see my branch debug-brain-connectivity-app to monitor which callbacks are executed)

Another way to look at this is the network activity in the dev tools, there are several big chunks. The first one is the layout loading, but then there are three other big chunks while in the covid app for example there is only one long dash_update_component call for the slicer. So I would first try to disable the initial calls and see if this improves stuff. I have to say it's not very easy to debug an app with 12 callbacks + the ones of the slicer!!

Finally maybe @nicolaskruchten would have ideas here!

image

surchs:
Thanks a lot @emmanuelle, I will look into these points! I don't think I can see your branch `debug-brain-connectivity-app` but for now I think I can do these locally.

emmanuelle:
@surchs sorry the push of my branch had not completed, now it's pushed for real :-)

emmanuelle (approved):

PR #543, opened 2020-12-21
DBC layout for covid x-ray app

surchs:
- condensed style-covid.css, removed style.css
- added navbar
- added modal with modal.md to explain app use and corresponding callback
- put layout elements in dbc cards
- reformatted hints and toolstips, put them inside the card footers
- removed annotation from histogram title and put them in the card footer
- moved conditional warning from histogram title to dbc toast popup and linked in callback

# App pull request
- [X] I am improving an existing app (redesigns/code "makeovers")

## About

- Playground deployment URL (new version): https://dash-playground.plotly.host/dash-covid-xray/
- Current gallery app URL: https://dash-gallery.plotly.host/dash-covid-xray/
- Python app repository link:


emmanuelle:
Thanks a ton for the various improvements ! The app looks beautiful now, the source and learn more buttons are very helpful, and so are the improved texts of the different cards. The app is maybe a little less compact than before with a bit more white space, on my laptop screen with the version on the gallery I used to decrease the zoom level at 80% and be able to see the 4 blocks on one screen. Is there an easy way to make the different cards more compact? I know it's the end of the project, so if there is no fast solution let's merge it like this. Thanks again!

surchs:
Oh yeah, I could use the full width and then make the cards height smaller. If I don't explicitly set the cards height to something smaller, you need to scroll to see the second row. But if we don't need the aspect ration of the viewers preserved, we could just fix it like that!

surchs:
I noticed a strange phenomenon that on first app load, the colored outlines of the slicer views seem to not really span the image: 
image

Not clear on where that's coming from. I think I'll make a slicer issue about this.

emmanuelle:
I also think it's a slicer issue, it appears on page load. Moving the sliders makes the problem disappear.

apps/dash-covid-xray/assets/style-covid.css
```diff
@@ -1,56 +1,18 @@
-/* make the graph still visible while the loading spinner is displayed */
-*[data-dash-is-loading="true"]{
-    visibility: visible;
+.card {
+    margin-top: 1em;
+    height: 50vh;
```
emmanuelle, line 3:
```suggestion
    height: 60vh;
```

emmanuelle: so that figures a bit bigger (I made some quick tests on my laptop and with 60 you can see the 4 figures on one screen, not all the text but for this you can scroll a bit).

emmanuelle (dismissed):

surchs:
ok, looks good to me with the new height!

emmanuelle (approved):

PR #560, opened 2021-02-11
Add automation to automatically release zipped apps

### LINKED ISSUE #559: Auto-release and upload individual apps as zip files
I experimented with github actions in [my personal repo](https://github.com/xhlulu/dash-sample-apps/blob/master/.github/workflows/main.yml) to automatically bump the version and release new versions of this repository every time a commit is pushed to `master` along **with the zip file of each app in `apps/`**.  I don't think the versioning adheres to semver, but by doing so it would be possible to download individual apps as a zip file instead of the entire repo (and history). 

For example:
```bash
wget https://github.com/plotly/dash-sample-apps/releases/latest/download/<appname>.zip
unzip <appname>.zip
cd <appname>/
pip install -r requirements.txt
python app.py
```

@alexcjohnson @nicolaskruchten thoughts on this?

xhlulu:
Closes #559

Whenever a commit is pushed to master:
- Automatically bump version
- Compress each app directory into a zip file
- Create a release and upload all the zipped apps

# App pull request
- [ ] This is a new app

## About

## Workflow

- [ ] I have created a branch in the appropriate monorepo, and the
  `assets/` folder.

## The pre-review review

- [ ] Does everything in my code serve some purpose? (I have removed
- [ ] Does everything in my code have a clear purpose? (My code is
- [ ] Am I reinventing the wheel? (I have used appropriate packages to

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 This is a monorepo designed to host all of the apps that have been
 created for the Python Dash Gallery.
 
-## Running an example app
+## Running an example app after cloning the repo
 
 You will need to run applications, and specify filenames, from the
 root directory of the repository. e.g., if the name of the app you
@@ -13,6 +13,20 @@ of the repository.
 
 Each app has a requirements.txt, install the dependecies in a virtual
 environment.
+
+## Downloading and running a single app
+
+If you don't want to clone the entire repo, you can also download an individual app as a zip file (see the [releases](https://github.com/plotly/dash-sample-apps/releases)), decompress and run them. In the example below, replace `<name>` by the name of the app:
+
+```bash
+wget https://github.com/plotly/dash-sample-apps/releases/latest/download/<name>.zip
+unzip <name>.zip
+cd <name>/
+python -m venv venv
+source venv/bin/activate  # Windows: \venv\scripts\activate
+pip install -r requirements.txt
+python app.py
+```
 
 ## Contributing to the sample apps repo
 
--- a/README.md
+++ b/README.md
@@ -16,15 +16,16 @@ environment.
 
 ## Downloading and running a single app
 
-If you don't want to clone the entire repo, you can also download an individual app as a zip file (see the [releases](https://github.com/plotly/dash-sample-apps/releases)), decompress and run them. In the example below, replace `<name>` by the name of the app:
+Visit the [releases page](https://github.com/plotly/dash-sample-apps/releases) and download and `unzip` the app you want. Then `cd` into the app directory and install its dependencies in a virtual environment in the following way:
 
 ```bash
-wget https://github.com/plotly/dash-sample-apps/releases/latest/download/<name>.zip
-unzip <name>.zip
-cd <name>/
 python -m venv venv
 source venv/bin/activate  # Windows: \venv\scripts\activate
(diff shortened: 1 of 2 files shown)


alexcjohnson:
What if we only ever had one release available? ie after uploading the new release we delete the previous one(s)? This feels like a bit of a misuse of the release mechanism, but it would make https://github.com/plotly/dash-sample-apps/releases much cleaner, and folks would likely just go there and download the zip they want.

xhlulu:
Yeah it shouldn't be too hard. 

I guess it does make it cleaner, although Github seems to automatically hide the assets of older releases to make it easier to navigate. 
image

Moreover replacing the version number with "latest" automatically resolves it. Though I agree it looks a lot cleaner when you look at the tags.

I think one potential advantage will be to more easily access older versions of a certain app, especially if we start versioning based on dash's major version (e.g. `dash-sample-apps v1.*` corresponds to `dash v1.*`, etc.) as we start refactoring older apps using new syntax (e.g. pattern matching). However this can still be done by simply cloning the entire repo, which shouldn't be too big of a problem once we refactor the git history and move the data files to `plotly/datasets`.

README.md
```diff
@@ -14,6 +14,20 @@ of the repository.
 Each app has a requirements.txt, install the dependecies in a virtual
 environment.
 
+## Downloading and running a single app
+
+If you don't want to clone the entire repo, you can also download an individual app as a zip file (see the [releases](https://github.com/plotly/dash-sample-apps/releases)), decompress and run them. In the example below, replace `<name>` by the name of the app:
+
+```bash
+wget https://github.com/plotly/dash-sample-apps/releases/latest/download/<name>.zip
```
alexcjohnson, line 22: Do we really want to document `wget` as the main entry point? I'd think a friendlier route would be from the website like:

---

Visit the [releases](https://github.com/plotly/dash-sample-apps/releases) page and download and unzip the app you want. Then `cd` into the app directory and install its dependencies in a virtual environment like:
```bash
python -m venv venv
source venv/bin/activate  # Windows: \venv\scripts\activate
pip install -r requirements.txt
```
Then run the app:
```bash
python app.py
```

---

separate note: wouldn't we want to suggest the same pattern above when cloning the whole repo? ie `cd` into the app dir, set up the virtual env in that dir, and run the app from there? The way we describe it now with running things from the repo root makes it unclear how to handle app-specific venvs.

alexcjohnson (dismissed):
💃 OK, let's keep it as you have, and we can revisit if the many releases become overwhelming. It feels a little confusing to me since most releases will only update one or two apps, but I agree that it could sometimes be useful to go back in time to an older version of an app.

Just a documentation comment, but this looks great!

xhlulu:
Thanks, I updated the readme with your suggestions; will merge now.

> separate note: wouldn't we want to suggest the same pattern above when cloning the whole repo? ie cd into the app dir, set up the virtual env in that dir, and run the app from there? The way we describe it now with running things from the repo root makes it unclear how to handle app-specific venvs.

I think it's a good idea. In fact I feel a lot of the readme might be confusing for the community; it feels to me it was written for internal use. Would be nice to refactor the entire readme and move some parts to contributing.md

PR #589, opened 2021-03-30
Propose various improvement to circleci workflow

xhlulu:
* Set `create_app=true` when deploying the gallery so we don't have to manually create the app beforehand
* Make `runtime.txt` optional during deployment
* Make the `assets` directory optional if the app does not contain any custom css or images
* Remake flake8 from tests, since black already covers most of what we want.


xhlulu:
@ycaokris Trying it out here: https://github.com/plotly/dash-sample-apps/tree/test-new-circleci

PR #629, opened 2021-05-10
Fix: add src=app.get_asset_url() to NOVA-IMS demo apps

AirballClaytonCalvin:


AirballClaytonCalvin:
Yup, totally. The [NOVA-IMS](https://dash-gallery.plotly.host/Portal/?search=nova) apps are student quality and are not constant with the Plotly look and feel. Once Dave has used them for sales purposes, they should either be taken down, or re-written.

PR #648, opened 2021-06-04
Optimize GIF sizes (#minor)

xhlulu:
Code used:

```bash
files=$(find apps/dashr* -type f -name '*.gif')

for file in $files; do
    gifsicle -O3 --colors 128 --resize-fit-height 800 -o $file $file
done
```


xhlulu:
Could use this python code to compress some of the csv files:
```python
import gzip
from glob import glob
import shutil
from tqdm.auto import tqdm

def convert_to_gz(file):
    with open(file, 'rb') as f_in:
        with gzip.open(file+'.gz', 'wb') as f_out:
            shutil.copyfileobj(f_in, f_out)

files = glob('**/*.csv', recursive=True)
len(files)

for file in tqdm(files):
    convert_to_gz(file)
```

PR #710, opened 2022-05-20
2022 refactor - removing-dash-r

danton267:
PR part of the 2022 refactor, bringing all Python apps up to date.


danton267 (commented):
removing the ramining r apps

